PR #221, opened 2024-04-15
fix: export defaults to environment variables

jta:
Setting a default in our env struct currently only affects the value being read into the struct. It is generally useful to have that value exporting back to the environment, in the case where other libraries source their configuration from the environment directly.

An existing usecase is where we modify the default for `AWS_MAX_ATTEMPTS`. The AWS SDK reads this environment variable directly when configuring the client.

### DIFF
--- a/handler/env.go
+++ b/handler/env.go
@@ -1,0 +1,29 @@
+package handler
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/sethvargo/go-envconfig"
+)
+
+// exportEnvVar ensures we export resolved values back into the environment
+// This is useful for cases where we set a default in our env struct, and need
+// the value to be propagated to other processes.
+func exportEnvVar(_ context.Context, _, resolvedKey, _, resolvedValue string) (newValue string, stop bool, err error) {
+	os.Setenv(resolvedKey, resolvedValue)
+	return resolvedValue, false, nil
+}
+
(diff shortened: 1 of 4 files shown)


handler/env.go
```diff
@@ -0,0 +1,29 @@
...
+func exportEnvVar(_ context.Context, _, resolvedKey, _, resolvedValue string) (newValue string, stop bool, err error) {
+	os.Setenv(resolvedKey, resolvedValue)
+	return resolvedValue, false, nil
+}
+
+// ProcessEnv populates struct from environment variables.
+func ProcessEnv(ctx context.Context, v any) error {
+	err := envconfig.ProcessWith(ctx, &envconfig.Config{
+		Target:   v,
+		Mutators: []envconfig.Mutator{envconfig.MutatorFunc(exportEnvVar)},
```
obs-gh-alexlew, line 23: so this'll end up taking all the resolved config values and re-exporting them all back to env?

obs-gh-alexlew (approved):

handler/env.go
jta: Yes. Unfortunately there's no way to tell if an environment variable has changed in the process. I was initially hopeful that I could compare `originalValue` vs `resolvedValue`, but in practice `envconfig` sets the default early enough that we can't disambiguate between the cases.

PR #223, opened 2024-04-17
feat(forwarder): set version on build

jta:
We want to set the version at build time. Unfortunately, the `sam build` command makes this slightly convoluted, given it executes `make build-${Resource}` without inheriting any of the variables passed into our original `make sam-build` call.

As a workaround, we stash the desired version in a file and source it back in. This is inherently race-y, since concurrent builds may overwite the file repeatedly. This risk should be fine for the time being, since we're always writing the same version value and we currently build serially.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ AWS_REGION ?= $(shell aws configure get region)
 SAM_BUILD_DIR ?= .aws-sam/build
 SAM_CONFIG_FILE ?= $(shell pwd)/samconfig.yaml
 SAM_CONFIG_ENV ?= default
+BUILD_MAKEFILE_ENV_VARS = .make.env
 
 DEBUG ?= 0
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -166,7 +168,7 @@ sam-publish-all:
 build-App:
 	$(call check_var,APP)
 	$(call check_var,ARTIFACTS_DIR)
-	GOARCH=arm64 GOOS=linux go build -tags lambda.norpc -o ./bootstrap cmd/$(APP)/main.go
+	GOARCH=arm64 GOOS=linux go build -tags lambda.norpc -ldflags "-X $(shell go list -m)/version.Version=${VERSION}" -o ./bootstrap cmd/$(APP)/main.go
```
obs-gh-alexlew, line 171: what does `shell go list -m` output?

obs-gh-alexlew (approved):

Makefile
jta: The package root, `github.com/observeinc/aws-sam-apps`

PR #226, opened 2024-04-19
feat(forwarder): handle records concurrently

jta:
Copying large objects can take a while when executing across regions. This commit ensures we handle each SQS record concurrently, which should be safe since:
- `s3:CopyObject` does not consume lambda resources, so we are not at risk of running out of memory / cpu
- our fan out factor is capped by our batch size on the SQS queue, which is quite small when compared to the AWS API limits.

Each SQS record itself may contain a list of files to copy. This list is still processed serially for now. In practice we do not use this capability yet.

### DIFF
--- a/handler/forwarder/handler.go
+++ b/handler/forwarder/handler.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -41,6 +42,7 @@ type Handler struct {
 	Override          Override
 	SourceBucketNames []string
 	Now               func() time.Time
+	MaxConcurrency    int
 }
 
 // GetCopyObjectInput constructs the input struct for CopyObject.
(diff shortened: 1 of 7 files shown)


handler/forwarder/handler.go
```diff
@@ -105,65 +107,91 @@ func (h *Handler) WriteSQS(ctx context.Context, r io.Reader) error {
...
+		sourceURL, err := url.Parse(copyRecord.URI)
+		if err != nil {
+			logger.Error(err, "error parsing source URI", "SourceURI", copyRecord.URI)
+			continue
 		}
-	}()
 
-	encoder := json.NewEncoder(&messages)
+		if !h.isBucketAllowed(sourceURL.Host) {
+			logger.Info("Received event from a bucket not in the allowed list; skipping", "bucket", sourceURL.Host)
```
obs-gh-alexlew, line 122: why logger.info here but logger.v(1).Info below?

obs-gh-alexlew (approved):

handler/forwarder/handler.go
jta: I just moved around the existing code, so not entirely sure. I assume that the rationale is that `V(0)` is used for a non-critical misconfiguration. You are subscribing a bucket, but did not grant the lambda permissions.

On the other hand, rejecting a file due to size is not a misconfiguration, it is the enforcement of a policy. There is no other way of preventing large files from triggering the lambda, since you can only subscribe based on prefix, and you may not control what is getting dumped into the bucket. As such it may warrant a higher verbosity.

PR #233, opened 2024-04-23
feat(stack): auto-subscribe to AWS Config

jta:
When enabled, AWS Config emits events to EventBridge. This opens the opportunity to ingest AWS Config data without having to manually subscribe an S3 bucket (for config history, snapshot and oversized notifications) and SNS Topic (for change notifications).

We add a new parameter, `ConfigDeliveryBucketName`, which gates whether to attempt to ingest config data. When present:

- we add the bucket to the source bucket list for our forwarder
- we add an eventbridge rule to construct copy commands for the forwarder on successful config delivery
- we add an eventbridge rule to forward all Config events

### DIFF
--- a/apps/stack/template.yaml
+++ b/apps/stack/template.yaml
@@ -258,6 +258,17 @@ Resources:
         TopicARN: !Ref Topic
         IncludeResourceTypes: !Join [",", !Ref IncludeResourceTypes]
         ExcludeResourceTypes: !Join [",", !Ref ExcludeResourceTypes]
+  ConfigSubscription:
+    Type: AWS::Serverless::Application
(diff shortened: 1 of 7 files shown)


apps/stack/template.yaml
```diff
@@ -79,12 +80,18 @@ Parameters:
       of this stack, and must therefore be unique within the account.
     Default: ""
     MaxLength: 51
+  ConfigDeliveryBucketName:
```
obs-gh-abhishekrao, line 83: Where do we manage IAM permissions for the forwarder for this pre-existing Config bucket?

obs-gh-abhishekrao (approved):

PR #265, opened 2024-05-17
fix(forwarder): adjust queue attributes

jta:
Match our defaults in terraform.


jta:
This turns out to be a bad idea. Increasing the batch window introduces a terrible delay in the case of low volume traffic, [as described here](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html):

```If you're using a batch window and your SQS queue contains very low traffic, Lambda might wait for up to 20 seconds before invoking your function. This is true even if you set a batch window lower than 20 seconds.```

PR #296, opened 2024-06-04
fix(config): adjust precedence of parameters.

jta:
The AWS Config API does not allow both exclusion and inclusion sets. We currently allow users to provide both, and prefer `ExcludeResourceTypes`, which can be misleading if the intent was to restrict collection further.

This commit changes precendence to prefer `IncludeResourceTypes` over `ExcludeResourceTypes`. The reasoning here is that the inclusion list tends to be small, in which case it is the cost conservative behavior.

An alternative approach would be to assert that only one of the sets is provided. Unfortunately errors in cloudformation are hard to debug, so the user experience would be terrible.

### DIFF
--- a/apps/config/template.yaml
+++ b/apps/config/template.yaml
@@ -109,16 +109,17 @@ Conditions:
       - ','
       - !Ref IncludeResourceTypes
   IncludeResources: !And
+    - !Condition EnableConfig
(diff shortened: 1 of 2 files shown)


NathanNam (approved):
LGTM

PR #324, opened 2024-07-29
fix(forwarder): improve CSV handling

jta:
A few simple performance fixes for converting CSV to JSON:
- avoid unmarshalling if target is json.RawMessage, since that calls `json.Valid`.
- avoid calling `strconv.Quote` where possible. Most of the inputs we handle in practice do not require special handling.
- keep track of max record length and use it to initialize our buffer, which avoids resizing.

In profiling, these changes reduced processing time for a VPC flow logs file by ~33%.


pkg/handler/forwarder/s3http/internal/decoders/csv.go
```diff
@@ -74,20 +75,35 @@ func (dec *CSVDecoder) Decode(v any) error {
...
 	}
 	buf.WriteString(`}`)
 
-	if err := json.Unmarshal(buf.Bytes(), v); err != nil {
+	if buf.Len() > dec.maxSize {
+		dec.maxSize = buf.Len()
+	}
+
+	// avoid unmarshalling if possible
+	if r, ok := v.(*json.RawMessage); ok {
```
obs-gh-alexlew, line 104: no additional tests for this new branch?

jta: This is actually the only case we test today, since both the tests and the runner use json.RawMessage. Arguably we could add a testcase for the unused path, but that was tested prior so we do know it works. We would be ensuring that it continues to work going forward. I've update the test to verify this slow path. The test results differ due to key reordering (when we go through `json.Unmarshal`, keys are ordered alphabetically)

obs-gh-alexlew (approved):

PR #334, opened 2024-08-28
fix: OB-35469 add subscription request parent span as child of disco…

obs-gh-virjramakrishnan:
…very request

### DIFF
--- a/pkg/handler/subscriber/instrument.go
+++ b/pkg/handler/subscriber/instrument.go
@@ -8,6 +8,8 @@ import (
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/aws/aws-sdk-go-v2/service/sqs"
 	"github.com/aws/aws-sdk-go-v2/service/sqs/types"
+	"github.com/observeinc/aws-sam-apps/pkg/tracing"
+
 	"github.com/go-logr/logr"
 	"go.opentelemetry.io/contrib/propagators/b3"
 	"go.opentelemetry.io/otel/codes"
@@ -25,8 +27,6 @@ func (h *InstrumentedHandler) HandleSQS(ctx context.Context, request events.SQSE
 	for _, record := range request.Records {
 		var req Request
 		var err error
-		logger.V(3).Info("Getting context from message attributes")
-		ctx = NewSQSCarrier().Extract(ctx, record.MessageAttributes)
 
 		if err = json.Unmarshal([]byte(record.Body), &req); err == nil {
 			_, err = h.HandleRequest(ctx, &req)
@@ -66,7 +66,7 @@ func (q *InstrumentedSQSClient) SendMessage(ctx context.Context, msg *sqs.SendMe
 	logger := logr.FromContextOrDiscard(ctx)
 	logger.V(3).Info("Injecting context into message attributes")
 	msg.MessageAttributes = make(map[string]types.MessageAttributeValue)
-	if err := NewSQSCarrier().Inject(ctx, msg.MessageAttributes); err != nil {
+	if err := tracing.NewSQSCarrier().Inject(ctx, msg.MessageAttributes); err != nil {
 		return nil, fmt.Errorf("failed to inject context into message attributes: %w", err)
 	}
 	logger.V(3).Info("sending message %s", msg.MessageBody)
--- a/pkg/lambda/subscriber/lambda.go
+++ b/pkg/lambda/subscriber/lambda.go
@@ -130,6 +130,6 @@ func New(ctx context.Context, cfg *Config) (*Lambda, error) {
 		return nil, fmt.Errorf("failed to register functions: %w", err)
 	}
 
-	l.Entrypoint = tracing.NewLambdaHandler(mux, tracerProvider)
+	l.Entrypoint = tracing.WrapHandlerSQSContext(tracing.NewLambdaHandler(mux, tracerProvider))
 	return l, nil
 }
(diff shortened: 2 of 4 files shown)


pkg/handler/subscriber/wrap_sqs_context.go
```diff
@@ -0,0 +1,47 @@
...
+	handler lambda.Handler
+}
+
+// Compile time check our Handler implements lambda.Handler.
+var _ lambda.Handler = SQSCheckerWrappedHandler{}
+
+// Invoke adds OTel span surrounding customer Handler invocation.
+func (h SQSCheckerWrappedHandler) Invoke(ctx context.Context, payload []byte) ([]byte, error) {
+
+	eventType := reflect.TypeOf(events.SQSEvent{})
```
obs-gh-alexlew, line 24: do we need to jump through all these hoops just to parse the sqs event?

pkg/handler/subscriber/wrap_sqs_context.go
```diff
@@ -0,0 +1,49 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambda"
+	"github.com/go-logr/logr"
+)
+
+type SQSCheckerWrappedHandler struct {
```
obs-gh-alexlew, line 13: why is this called SQSChecker? what is it checking?

obs-gh-virjramakrishnan: It checks if the payload is a SQS event and if so, injects context, I suppose SQSContextInjector might be more appropriate.

obs-gh-alexlew: Its not really injecting the context, its parsing it out right? Something like `SQSWithContextHander`

obs-gh-virjramakrishnan: Perhaps I am misunderstanding how "Inject" is used in this context - what I meant was that we are using the context provided by the SQS event. But I like `SQSWithContextHander`

obs-gh-virjramakrishnan: did you mean `SQSWithContextHandler`

pkg/handler/subscriber/instrument.go
```diff
@@ -25,8 +25,6 @@ func (h *InstrumentedHandler) HandleSQS(ctx context.Context, request events.SQSE
 	for _, record := range request.Records {
 		var req Request
 		var err error
-		logger.V(3).Info("Getting context from message attributes")
-		ctx = NewSQSCarrier().Extract(ctx, record.MessageAttributes)
```
jta, line 29: This is suspicious. If you remove this, what functionality is `InstrumentedHandler` providing?

jta (commented):
It's hard to tell what the fix is exactly. Ideally you'd provide a git commit message that explains why things weren't working before, and why your change fixes it.

pkg/handler/subscriber/instrument.go
obs-gh-virjramakrishnan: I see your point. This effectively just becomes a wrapper around `HandleRequest` that does nothing. There might be some more refactoring that is possible if I do remove this. The idea with this modification is to change the place where the context is being inserted from. I removed adding the context in here because I am adding it in in a different place, as a wrapper to the invocation. This allows the parent span of the subscription request to get the context from the SQS event (the context of the previous request). I'll add more description in the commit message.

pkg/handler/mux.go
```diff
@@ -90,10 +95,15 @@ func (m *Mux) Invoke(ctx context.Context, payload []byte) (response []byte, err
 	}
 	ctx = logr.NewContext(ctx, logger)
 
+	span := trace.SpanFromContext(ctx)
```
obs-gh-alexlew, line 98: this seems more tracing related than handler/mux related? did this have to move here?

obs-gh-virjramakrishnan: Ah so the reason I moved it here is that the Otel wrapper directly calls this - if we want that span to show up as in error, and have the payload show up in the we need to get the context after the Otel wrapper span has modified the context (otherwise there will be no span associated with ctx, and the otel will create a span and will would add the payload and error handling to this new blank span). I therefore factored the code into the mux handler.

obs-gh-virjramakrishnan:
Hi @jta, just to clarify what the bug was: 

Before, the OTEL created traces were being created, and _then_, we would extract context from the SQS message if we received one, and replace the context. However, the effect of this is that all remaining spans in the Subscription request would be children of the HandleRequest from the discovery request, as opposed to the subscriber request. This meant that the subscriber spans were left childless. The "missing spans" were actually under the discovery request. To fix this, I made another wrapper that wraps around the OTEL wrapper. This means that we extract the context _before_ we take start the otel spans - the context we pass in is the one from the discovery request, and so all of the subscription spans show up as children of the discovery handleRequest span.

pkg/lambda/subscriber/lambda.go
```diff
@@ -130,6 +130,6 @@ func New(ctx context.Context, cfg *Config) (*Lambda, error) {
 		return nil, fmt.Errorf("failed to register functions: %w", err)
 	}
 
-	l.Entrypoint = tracing.NewLambdaHandler(mux, tracerProvider)
+	l.Entrypoint = tracing.WrapHandlerSQSContext(tracing.NewLambdaHandler(mux, tracerProvider))
```
obs-gh-alexlew, line 133: new wrapper that only applies in the subscriber case since in forwarder we're not injecting the context into the sqs messages/trying to read it back out.

PR #338, opened 2024-09-20
feat: OB-36691 add lambda and s3 bucket to record push metrics config

obs-gh-virjramakrishnan:


.DS_Store
obs-gh-alexlew, line 1: this shouldn't be in the diff should it?

DEVELOPER.md
```diff
@@ -37,7 +37,7 @@ and the most important variables, run `make help`, e.g:
 VARIABLES:
   APPS          = config configsubscription firehose forwarder logwriter metricstream stack
   AWS_REGION    = us-west-2
-  GO_BINS       = forwarder subscriber
+  GO_BINS       = forwarder subscriber metricsrecorder
```
obs-gh-alexlew, line 40: what metrics is this new app recording? really it updates the existing metrics stream right? maybe we could call it `metricstreamupdater` or `metricsconfigurator`

apps/metricstream/template.yaml
```diff
@@ -27,8 +28,8 @@ Parameters:
     Type: String
     Description: >-
       A file hosted in S3 containing list of metrics to stream.
-    Default: 's3://observeinc/cloudwatchmetrics/filters/recommended.yaml'
-    AllowedPattern: "^s3:\/\/.*"
+    Default: 's3://observeinc/cloudwatchmetrics/filters/empty.yaml'
+    AllowedPattern: "^(s3:\/\/.*)?$"
```
obs-gh-alexlew, line 32: why is the pattern changing?

apps/metricstream/template.yaml
```diff
@@ -61,11 +62,33 @@ Parameters:
     Description: |
       Buffer incoming data to the specified size, in MiBs, before delivering it
       to the destination.
-
+  ObserveAccountID:
```
obs-gh-alexlew, line 65: do we not need the cluster too? since the urls are different based on the cluster

apps/metricstream/template.yaml
```diff
@@ -61,11 +62,33 @@ Parameters:
...
+    Type: String
+    Description: Observe Account Number
+    AllowedPattern: \d*
+  Timestamp:
+    Type: String
+    Description: Timestamp for the metric stream.
+    Default: ''
+    AllowedPattern: "^[0-9]*$"
+  DatasourceID:
```
obs-gh-alexlew, line 74: why do we need the datasourceId?

.DS_Store
obs-gh-virjramakrishnan: Oh shoot sorry, not sure why that wasn't gitignored, mb

apps/metricstream/template.yaml
```diff
@@ -188,7 +224,81 @@ Resources:
           Location: !Ref FilterUri
       # end of processable content for include
       OutputFormat: !Ref OutputFormat
-
+  MetricsRecorderExecutionRole:
```
obs-gh-alexlew, line 227: think its fine to just call it `AppNameRole` ex. `MetricsConfiguratorRole`

apps/metricstream/template.yaml
```diff
@@ -188,7 +224,81 @@ Resources:
...
+      Handler: bootstrap
+      Runtime: provided.al2
+      Architectures:
+        - arm64
+      Environment:
+        Variables:
+          VERBOSITY: 6
+          ACCOUNT_NUMBER: !Ref ObserveAccountID
+          DATASOURCE_ID: !Ref DatasourceID
+          SECRET_NAME: !Ref GQLTokenSecret
```
obs-gh-alexlew, line 282: this is the key of the secret, not the name right?

pkg/handler/metricsrecorder/handler.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+type MetricsListItem struct {
+	Namespace   string   `json:"namespace"`
+	MetricNames []string `json:"metricNames"`
+}
+
+func convertToMetricStreamFilters(MetricsList []MetricsListItem) []types.MetricStreamFilter {
+
+	if len(MetricsList) == 0 {
```
obs-gh-alexlew, line 55: dont think we need to handle the 0 care do we? just initalize the slice as empty and then the range will just not run if its empty

apps/metricstream/template.yaml
obs-gh-virjramakrishnan: Since there are two options for how to configure metrics (via config file or via lambda), it is feasible that a user does not have a file to use. the first pattern forces a s3 file to be inserted. While we could have a parameter as a default, I think we should be flexible to someone entering nothing here (since we don't actually need this in the case of configuration via lambda).

pkg/handler/metricsrecorder/handler.go
obs-gh-virjramakrishnan: Agreed, there was some weird error with the types being correct, and this was my strong-armed way of fixing it, but your way is more clear.

pkg/handler/metricsrecorder/handler.go
```diff
@@ -0,0 +1,338 @@
...
+	jsonData, err := json.Marshal(gqlRequest)
+	if err != nil {
+		report_err := h.reportStatus(req, false, fmt.Sprintf("failed to marshall data into gql request %s", err))
+		if report_err != nil {
+			return nil, fmt.Errorf("failed to report status to cloudformation: %w, for error %w", report_err, err)
+		}
+		return nil, fmt.Errorf("failed to marshal request into gql request: %w", err)
+	}
+
+	host := fmt.Sprintf("%s.observe-eng.com", accountNumber)
```
obs-gh-alexlew, line 201: this won't always be observe-eng.com

apps/metricstream/template.yaml
obs-gh-virjramakrishnan: In effect, yes. AWS Secretsmanger calls keys 'secret names', and the value of the secret 'values'

pkg/handler/metricsrecorder/handler.go
```diff
@@ -0,0 +1,338 @@
...
+			return nil, fmt.Errorf("failed to report status to cloudformation: %w, for error %w", report_err, err)
+		}
+		return nil, fmt.Errorf("error reading response body: %w", err)
+	}
+
+	var result GraphQLResponse
+	err = json.Unmarshal(bodyBytes, &result)
+	if err != nil {
+		report_err := h.reportStatus(req, false, fmt.Sprintf("error unmarshalling JSON from GQL response %s", err))
+		if report_err != nil {
```
obs-gh-alexlew, line 240: can we get a helper function for trying to report status to cloudformation?

pkg/handler/metricsrecorder/handler.go
```diff
@@ -0,0 +1,338 @@
...
+	hasMetricsField := false
+	for _, variable := range variablesList {
+		if variable.Name == "Metrics" {
+			hasMetricsField = true
+			break
+		}
+	}
+
+	if !hasMetricsField {
+		report_err := h.reportStatus(req, false, "no metrics field in response")
```
obs-gh-alexlew, line 261: "metrics variable not set in datasource"

pkg/handler/metricsrecorder/handler.go
```diff
@@ -0,0 +1,338 @@
...
+		if report_err != nil {
+			return nil, fmt.Errorf("failed to report status to cloudformation: %w, for error %w", report_err, err)
+		}
+		return nil, fmt.Errorf("no metrics field in response")
+	}
+
+	if len(result.Data.Datasource.Variables) > 0 {
```
obs-gh-alexlew, line 268: confused here, why are we checking for any variables? we just checked above for a specific one? also why are we just assuming the first one is the metrics one?

pkg/handler/metricsrecorder/handler.go
```diff
@@ -0,0 +1,338 @@
...
+		AccountNumber:    cfg.AccountNumber,
+		SecretName:       cfg.SecretName,
+		DatasourceID:     cfg.DatasourceID,
+		FirehoseArn:      cfg.FirehoseArn,
+		RoleArn:          cfg.RoleArn,
+		OutputFormat:     cfg.OutputFormat,
+	}, nil
+}
+
+func (h Handler) Invoke(ctx context.Context, payload []byte) ([]byte, error) {
```
obs-gh-alexlew, line 100: this function is a monster, i'd split it up into a couple smaller functions

DEVELOPER.md
```diff
@@ -37,7 +37,7 @@ and the most important variables, run `make help`, e.g:
 VARIABLES:
   APPS          = config configsubscription firehose forwarder logwriter metricstream stack
   AWS_REGION    = us-west-2
-  GO_BINS       = forwarder subscriber
+  GO_BINS       = forwarder subscriber metricsconfigurer
```
obs-gh-alexlew, line 40: i think configurer and configurator both work here, i just feel like configurator sounds better/is easier to say haha

apps/metricstream/template.yaml
obs-gh-alexlew: There's no way to specify that this param can be left blank? And the allowedPattern check still runs in that case?

apps/metricstream/template.yaml
```diff
@@ -61,11 +62,37 @@ Parameters:
...
+    Type: String
+    Description: Timestamp for the metric stream.
+    Default: ''
+    AllowedPattern: "^[0-9]*$"
+  DatasourceID:
+    Type: String
+    Description: >-
+      The datastream for this metric stream.
+    AllowedPattern: '\d*'
+  DomainName:
```
obs-gh-alexlew, line 79: nit: I'd call this ObserveDomainName and put it under the `ObserveAcountID` param

apps/metricstream/template.yaml
```diff
@@ -61,11 +62,37 @@ Parameters:
...
+    NoEcho: true
+    Description: >-
+      The token used to retrieve metric configuration.
 Conditions:
   UseStackName: !Equals
     - !Ref NameOverride
     - ''
+  DeployLambda: !Not
+    - !Equals
+      - !Ref DatasourceID
```
obs-gh-alexlew, line 94: wont the frontend always provide the datasourceID?

pkg/handler/metricsconfigurer/handler.go
```diff
@@ -0,0 +1,374 @@
...
+	logger := h.Logger
+	logger.V(4).Info("handling request")
+
+	req, err := h.parsePayload(payload)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse payload: %w", err)
+	}
+
+	// Handle Delete case, directly delete
+	if req.RequestType == "Delete" {
```
obs-gh-alexlew, line 102: where do these Delete requests come from?

pkg/handler/metricsconfigurer/handler.go
```diff
@@ -0,0 +1,374 @@
...
+		report_err := h.reportStatus(*req, false, "failed to load AWS config")
+		if report_err != nil {
+			return nil, fmt.Errorf("failed to report status to cloudformation: %w, for error %w", report_err, err)
+		}
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
+	token, err := h.getSecretValue(ctx, req, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve secret value: %w", err)
```
obs-gh-alexlew, line 121: kind of confused, why do we only do `h.reportStatus` sometimes on errors?

pkg/handler/metricsconfigurer/handler.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	logger.V(4).Info("returned response to cloudformation")
+	return []byte{}, nil
+}
+
+func (h Handler) parsePayload(payload []byte) (*Request, error) {
+	dec := json.NewDecoder(bytes.NewReader(payload))
+
+	var rawResult map[string]interface{}
```
obs-gh-alexlew, line 167: do we not know the schema of this payload?

pkg/handler/metricsconfigurer/handler.go
```diff
@@ -0,0 +1,374 @@
...
+func (h Handler) getSecretValue(ctx context.Context, req *Request, cfg aws.Config) (*string, error) {
+
+	secretName := h.SecretName
+	svc := secretsmanager.NewFromConfig(cfg)
+	secretValue, err := svc.GetSecretValue(ctx, &secretsmanager.GetSecretValueInput{
+		SecretId: &secretName,
+	})
+	if err != nil {
+		report_err := h.reportStatus(*req, false, fmt.Sprintf("failed to retrieve secret value %s", err))
+		if report_err != nil {
```
obs-gh-alexlew, line 255: mentioned this before, seems like we can just wrap these 4 lines into a helper function, they're fairly repetitive and it can just take in the specific error message as a param

pkg/handler/metricsconfigurer/handler.go
```diff
@@ -0,0 +1,374 @@
...
+
+	return &token, nil
+}
+
+func (h Handler) getDatasource(req *Request, token *string, domainName string, client *http.Client) ([]byte, error) {
+	logger := h.Logger
+	logger.V(4).Info("calling observe api", "accountNumber", h.AccountNumber)
+
+	fullToken := fmt.Sprintf("Bearer %s %s", h.AccountNumber, *token)
+	query := fmt.Sprintf(`
```
obs-gh-alexlew, line 271: this should probably be defined as a const instead of just in this line

pkg/handler/metricsconfigurer/handler_test.go
```diff
@@ -0,0 +1,390 @@
...
+                    "data": {
+                        "datasource": {
+                            "name": "TestDatasource",
+                            "variables": [
+                                {
+                                    "name": "Metrics",
+                                    "details": {
+                                        "awsCollectionStackConfig": {
+                                            "awsServiceMetricsList": [
+                                                {
```
obs-gh-alexlew, line 336: might be good to include multiple metrics here to make sure we're mapping over them all

static/cloudwatchmetrics/filters/empty.yaml
```diff
@@ -0,0 +1,2 @@
+IncludeFilters:
+  - Namespace: "None"
```
obs-gh-alexlew, line 2: is this an AWS level constant/sentinel value?

apps/metricstream/template.yaml
```diff
@@ -61,11 +62,41 @@ Parameters:
     Description: |
       Buffer incoming data to the specified size, in MiBs, before delivering it
       to the destination.
-
+  ObserveAccountID:
+    Type: String
+    Description: Observe Account Number
```
obs-gh-alexlew, line 67: nit: name is AccountId but this says Account Number, just stick to ID

apps/metricstream/template.yaml
```diff
@@ -61,11 +62,41 @@ Parameters:
...
+    AllowedPattern: '\d*'
+    Default: ''
+  ObserveDomainName:
+    Type: String
+    Description: >-
+      The domain name this request is coming from.
+    Default: ''
+  UpdateTimestamp:
+    Type: String
+    Description: Timestamp when the metric stream was created or updated
```
obs-gh-alexlew, line 77: is this a unix timestamp? nanoseconds?

apps/stack/template.yaml
```diff
@@ -310,7 +346,15 @@ Resources:
         - !Ref Topic
       Parameters:
         BucketArn: !GetAtt Bucket.Arn
-        FilterUri: !Ref MetricStreamFilterUri
+        FilterUri: !If
+          - DeployLambda
+          - 's3://observeinc/cloudwatchmetrics/filters/default.yaml'
```
obs-gh-alexlew, line 351: does the lambda ever use this file? if not does this even need to be a real s3 file?

obs-gh-alexlew: also we should write some comment here to explain why this is necessary

pkg/handler/metricsconfigurator/handler.go
```diff
@@ -0,0 +1,342 @@
...
+			Namespace:   &service.Namespace,
+			MetricNames: service.MetricNames,
+		})
+	}
+
+	return metricsFilters
+}
+
+func (h Handler) getSecretValue(ctx context.Context, cfg aws.Config) (*string, error) {
+
```
obs-gh-alexlew, line 227: nit: there seems to be a lot of whitespace right after the function definition line? rm

obs-gh-alexlew:
any unit tests around lambda vs. path in the actual app parameters and how that ends up affecting the result?

apps/metricstream/template.yaml
```diff
@@ -186,9 +238,101 @@ Resources:
         Name: "AWS::Include"
         Parameters:
           Location: !Ref FilterUri
-      # end of processable content for include
+      # End of processable content for include.
+      # The include function here is a macro that
+      # will be replaced with the contents of the file
+      # specified by the FilterUri parameter.
+      # The filterURI parameter is a decided in the
```
obs-gh-alexlew, line 245: is decided

apps/metricstream/template.yaml
```diff
@@ -186,9 +238,101 @@ Resources:
...
+      # parent stack, where, depending on whether we are
+      # going with a lambda based or yaml file based
+      # approach, we will use a different file.
+      # The reason this if statement is not here, in the
```
obs-gh-alexlew, line 249: nit: no comma needed

apps/metricstream/template.yaml
```diff
@@ -186,9 +238,101 @@ Resources:
...
+      # location parameter is because AWS does not allow
+      # to put an !If function inside an "AWS::Include"
```
obs-gh-alexlew, line 251: does not allow an !If function inside an "AWS::Include" block

apps/metricstream/template.yaml
```diff
@@ -186,9 +238,101 @@ Resources:
         Name: "AWS::Include"
         Parameters:
           Location: !Ref FilterUri
-      # end of processable content for include
+      # End of processable content for include.
+      # The include function here is a macro that
+      # will be replaced with the contents of the file
+      # specified by the FilterUri parameter.
+      # The filterURI parameter is a decided in the
+      # parent stack, where, depending on whether we are
```
obs-gh-alexlew, line 246: can you install this AWS Sam app separately without installing the whole stack? you should be able to

apps/stack/template.yaml
```diff
@@ -310,7 +362,26 @@ Resources:
...
+          - !Ref MetricStreamFilterUri
+          # If the user is using a lambda based approach, we will
+          # use a default metric filter. Not that this metric filter
+          # is different from the filter used for `empty.yaml`,
+          # which is a deliberate choice to avoid an error. If the
+          # files were to be the same, upon switching back from
+          # the lambda based approach to a yaml based approach,
+          # the cloudformation would otherwise detect no difference
+          # in the template-specified metric streams, thus no
+          # update would occur (and the metrics selected via lammbda would
```
obs-gh-alexlew, line 377: typo: lambda

apps/stack/template.yaml
```diff
@@ -310,7 +362,26 @@ Resources:
...
       Parameters:
         BucketArn: !GetAtt Bucket.Arn
-        FilterUri: !Ref MetricStreamFilterUri
+        FilterUri: !If
+          - DeployLambda
+          - 's3://observeinc/cloudwatchmetrics/filters/default.yaml'
+          - !Ref MetricStreamFilterUri
+          # If the user is using a lambda based approach, we will
+          # use a default metric filter. Not that this metric filter
+          # is different from the filter used for `empty.yaml`,
```
obs-gh-alexlew, line 371: need to mention that this file needs to exist as well

apps/stack/template.yaml
```diff
@@ -310,7 +359,19 @@ Resources:
         - !Ref Topic
       Parameters:
         BucketArn: !GetAtt Bucket.Arn
-        FilterUri: !Ref MetricStreamFilterUri
+        FilterUri: !If
+          - DeployLambda
+          - 's3://observeinc/cloudwatchmetrics/filters/default.yaml'
+          - !Ref MetricStreamFilterUri
+          # default.yaml must exist. When using a lambda for config we use
```
obs-gh-alexlew, line 366: # We provide a yaml file URI here where the contents are a sentinel value that the user would never enter.
# Then the lambda modifies the metricstream to reflect the actual config provided.
# We need this sentinel value here initially so that if we switch back to the yaml based approach,
# the cloudformation stack will recognize that the value from the yaml file has changed and will update.

obs-gh-alexlew (approved):

PR #342, opened 2024-10-03
feat: OB-37376 merge cf stacks for push and pull by adding poller rol…

obs-gh-virjramakrishnan:
…e to stack

### DIFF
--- a/apps/stack/template.yaml
+++ b/apps/stack/template.yaml
@@ -43,6 +43,12 @@ Metadata:
           - GQLToken
           - UpdateTimestamp
       - Label:
+          default: Enable Observe Metrics Poller
+        Parameters:
+          - MetricsPollerAllowedActions
+          - ObserveAwsAccountId
+          - DatastreamIds
+      - Label:
           default: Forwarder Options
         Parameters:
           - SourceBucketNames
@@ -158,6 +164,24 @@ Parameters:
     Description: >-
       The token used to retrieve metric configuration.
     Default: ''
+  MetricsPollerAllowedActions:
+    Type: CommaDelimitedList
+    Description: >-
+      List of actions allowed for the metrics poller
+      role. Leave empty to use default actions.
+    Default: cloudwatch:GetMetricData,cloudwatch:ListMetrics,tag:GetResources
+  ObserveAwsAccountId:
+    Type: String
+    Description: Observe AWS Account ID which will be allowed to assume role.
+    AllowedPattern: '\d*'
+    Default: ''
+  DatastreamIds:
+    Type: CommaDelimitedList
+    Description: Datastream IDs where data will be ingested to.
+      This ensures Observe cannot assume this role outside
+      of this context.
+    AllowedPattern: \d*
+    Default: ''
   SourceBucketNames:
     Type: CommaDelimitedList
     Description: >-
@@ -237,6 +261,10 @@ Conditions:
   EnableSubscription: !Or
     - !Condition HasLogGroupNamePatterns
     - !Condition HasLogGroupNamePrefixes
+  EnableMetricsPollerRole: !Not
+    - !Equals
+      - !Ref ObserveAwsAccountId
+      - ""
 Resources:
   Topic:
     Type: "AWS::SNS::Topic"
@@ -400,6 +428,21 @@ Resources:
           - UseStackName
           - !Sub "${AWS::StackName}-MetricStream"
           - !Sub "${NameOverride}-MetricStream"
+  MetricsPollerRole:
+    Type: AWS::Serverless::Application
+    Condition: EnableMetricsPollerRole
+    Properties:
+      Location: ../externalrole/template.yaml
+      NotificationARNs:
+        - !Ref Topic
+      Parameters:
+        ObserveAwsAccountId: !Ref ObserveAwsAccountId
+        AllowedActions: !Join [",", !Ref MetricsPollerAllowedActions]
+        DatastreamIds: !Join [",", !Ref DatastreamIds]
+        NameOverride: !If
+          - UseStackName
+          - !Sub "${AWS::StackName}-metrics-poller"
+          - !Sub "${NameOverride}-metrics-poller"
 Outputs:
   BucketName:
     Description: >-
@@ -488,3 +531,12 @@ Outputs:
     Value: !GetAtt
       - LogWriter
       - Outputs.SubscriberLogGroupName
+  PollerRoleArn:
+    Description: >-
+      ARN for IAM Role to be assumed by CloudWatch for metric delivery. This
+      value is required when configuring a subscription towards the Firehose
+      Delivery Stream.
+    Condition: EnableMetricsPollerRole
+    Value: !GetAtt
+      - MetricsPollerRole
+      - Outputs.RoleArn


apps/stack/template.yaml
```diff
@@ -37,6 +37,12 @@ Metadata:
           default: CloudWatch Metrics
         Parameters:
           - MetricStreamFilterUri
+      - Label:
+          default: Cloudwatch Metrics Poller
+        Parameters:
+          - ObserveAwsAccountId
+          - AllowedActions
```
obs-gh-alexlew, line 44: why are we taking this in as a param? shouldn't we know what we need?

apps/stack/template.yaml
```diff
@@ -110,6 +116,22 @@ Parameters:
...
     Default: 's3://observeinc/cloudwatchmetrics/filters/recommended.yaml'
     AllowedPattern: "^(s3:\/\/.*)?$"
+  ObserveAwsAccountId:
+    Type: String
+    Description: Observe AWS Account ID which will be allowed to assume role.
+    AllowedPattern: \d*
+  AllowedActions:
+    Type: CommaDelimitedList
+    Description: IAM actions that Observe account is allowed to execute.
+  DatastreamIds:
```
obs-gh-alexlew, line 126: is this check going to happen on the poller side? its not happening in the sam app side right?

apps/stack/template.yaml
obs-gh-virjramakrishnan: great point, let me fix that.

apps/stack/template.yaml
obs-gh-virjramakrishnan: are you referring to the validity of the ObserveAwsAccountId? I thought this would happen based on the frontend passing in correct input. I'll hardcode AllowedActions

apps/stack/template.yaml
```diff
@@ -378,6 +398,22 @@ Resources:
           - UseStackName
           - !Sub "${AWS::StackName}-MetricStream"
           - !Sub "${NameOverride}-MetricStream"
+  MetricsPoller:
```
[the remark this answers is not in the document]
obs-gh-virjramakrishnan: Sorry, I think this is a naming thing - what we are creating here is just the role and policy. Will update to `MetricsPollerRole` to reflect.

apps/stack/template.yaml
obs-gh-virjramakrishnan: This is because of how AWS recommends allowing other accounts to access your resources: see [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html). It seems that it is possible to do this process without having an external id, but it is not recommended. This is also how the poller works currently - we pass in the datastreamIds ([terraform](https://github.com/observeinc/terraform-observe-aws-quickstart/blob/58a5f29435d9ec0d51b6490fc29d77931d5919ac/sources/cwmetricscraper/main.tf#L18)) and then we use it to assume roles. We take it as input [here](https://gerrit.observeinc.com/plugins/gitiles/observe/+/refs/heads/master/code/go/src/observe/ingest/poller/cloudwatchmetrics/poller.go#74) and use it to get credentials we can use in our polling. So the stack here is just checking that the request is coming from our backend.

apps/stack/template.yaml
```diff
@@ -378,6 +398,22 @@ Resources:
...
+    Type: AWS::Serverless::Application
```
obs-gh-alexlew, line 402: its clearly not just a role and a policy, its a lambda based on this type?

obs-gh-alexlew: or the role and policy is bundled under this application?

obs-gh-virjramakrishnan: Not sure what you mean by already - the nested stack `externalrole` is only deployed if `ObserveAwsAccountId` is provided. Yes, it is just a role and policy - check out `externalrole/template.yaml`. Note that this nested stack existed in the aws-sam-apps repo before I made this commit. My change was that I made it such that the externalrole stack was conditionally nested under the main stack for push based, if `ObserveAwsAccountId` is provided. I didn't have to change `externalrole.yaml` at all.

apps/stack/template.yaml
```diff
@@ -378,6 +398,22 @@ Resources:
...
+      NotificationARNs:
+        - !Ref Topic
+      Parameters:
+        ObserveAwsAccountId: !Ref ObserveAwsAccountId
+        AllowedActions: >-
+          cloudwatch:GetMetricData,cloudwatch:ListMetrics,tag:GetResources
+        DatastreamIds: !Join [",", !Ref DatastreamIds]
+        NameOverride: !If
+          - UseStackName
+          - !Sub "${AWS::StackName}-MetricsPollerRole"
```
jta, line 415: Beware that this dictates the name of the IAM role, which in turn must match the ARN the Observe poller was created with. Any mismatch in name will lead to traffic being ignored.

I believe that if you just pass `NameOverride` as is there will not be any clash with other IAM roles, but worth double checking.

apps/stack/template.yaml
```diff
@@ -378,6 +398,22 @@ Resources:
           - UseStackName
           - !Sub "${AWS::StackName}-MetricStream"
           - !Sub "${NameOverride}-MetricStream"
+  MetricsPollerRole:
+    Type: AWS::Serverless::Application
+    Condition: EnableMetricsPollerRole
+    Properties:
+      Location: ../externalrole/template.yaml
```
jta, line 405: You should surface the outputs of this template in the parent stack.

apps/stack/template.yaml
```diff
@@ -378,6 +398,22 @@ Resources:
...
+      NotificationARNs:
+        - !Ref Topic
+      Parameters:
+        ObserveAwsAccountId: !Ref ObserveAwsAccountId
+        AllowedActions: >-
+          cloudwatch:GetMetricData,cloudwatch:ListMetrics,tag:GetResources
```
jta, line 411: You might want to parameterize `AllowedActions` and either default to this or leave it up to caller to define what actions are allowed.

apps/stack/template.yaml
```diff
@@ -37,6 +37,11 @@ Metadata:
           default: CloudWatch Metrics
         Parameters:
           - MetricStreamFilterUri
+      - Label:
+          default: Cloudwatch Metrics Poller Role
```
jta, line 41: This functionality may be useful beyond metrics collection, so maybe keep it as "Enable Observe poller"

apps/stack/template.yaml
obs-gh-virjramakrishnan: I want to avoid a clash between the metrics and resource poller, so I'm adding `-metrics-poller` as the suffix here and also in the place where the poller is created.

apps/stack/template.yaml
obs-gh-virjramakrishnan: I like the idea of making it a default - I'll do that!

apps/stack/template.yaml
obs-gh-virjramakrishnan: Sounds good!

apps/stack/template.yaml
```diff
@@ -42,6 +42,12 @@ Metadata:
           - DatasourceID
           - GQLToken
           - UpdateTimestamp
+      - Label:
+          default: Enable Observe Poller
```
obs-gh-alexlew, line 46: I know Joao had you switch it back to this but we're going to add another Label + parameter group here later for config poller right? And that'll be separate? It actually seems like we would want to differentiate between the two pollers vs just calling this Observe Poller

obs-gh-virjramakrishnan: My idea was that we could just integrate the roles necessary for both pollers into one role/policy. I agree that for now the best term is `Metrics Poller`. When the next poller comes in, I thought that we could choose to either separate it out/call it `Enable Observe Pollers`

obs-gh-alexlew: got it, so actually it won't be a separate label/parameter group for config pollers?

obs-gh-alexlew (approved):

apps/stack/template.yaml
obs-gh-virjramakrishnan: Honestly I haven't scoped out the amount of new roles that would be required, but I suspect it would be pretty limited, like some AllowReadResources or something like that, my thought is that's not enough to justify a whole other section.

PR #349, opened 2024-10-24
fix: OB-38134 make parent stack outputs conditioned on nested stack

obs-gh-virjramakrishnan:
fix to correct logs to work when subscriber is not enabled

### DIFF
--- a/apps/stack/template.yaml
+++ b/apps/stack/template.yaml
@@ -220,7 +220,21 @@ Conditions:
     - !Equals
       - !Ref DatasourceID
       - ""
-
+  HasLogGroupNamePatterns: !Not
+    - !Equals
+      - !Join
+        - ','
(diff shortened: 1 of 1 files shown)


apps/stack/template.yaml
```diff
@@ -220,7 +220,21 @@ Conditions:
     - !Equals
       - !Ref DatasourceID
       - ""
-
+  HasLogGroupNamePatterns: !Not
```
obs-gh-alexlew, line 223: Is there a * sentinel value that's valid for this?

obs-gh-virjramakrishnan: You can enter * for `LogGroupNamePatterns` and it will collect everything, yes

obs-gh-alexlew (approved):

PR #366, opened 2024-12-05
docs: clarify when filterUri is overridden

obs-gh-virjramakrishnan:
This makes updates to the docs, and also the comment under the input fields in the cloudformation stack creation page.


NathanNam (approved):
LGTM

obs-gh-alexlew (approved):

PR #369, opened 2024-12-09
feat: OB-39910 add multiple metric streams to work around 1000 metric…

obs-gh-virjramakrishnan:
… limit

### DIFF
--- a/apps/metricstream/template.yaml
+++ b/apps/metricstream/template.yaml
@@ -97,6 +97,8 @@ Conditions:
     - !Equals
       - !Ref DatasourceID
       - ""
+  NotDeployLambda: !Not
+    - !Condition DeployLambda
 
 Resources:
   DeliveryStreamRole:
@@ -224,6 +226,7 @@ Resources:
                 Resource: !GetAtt 'DeliveryStream.Arn'
   MetricStream:
     Type: AWS::CloudWatch::MetricStream
+    Condition: NotDeployLambda
     Metadata:
       cfn-lint:
         config:
@@ -264,7 +267,8 @@ Resources:
               - Effect: Allow
                 Action:
                   - cloudwatch:PutMetricStream
-                Resource: !GetAtt MetricStream.Arn
+                Resource: !Sub |-
+                  arn:aws:cloudwatch:${AWS::Region}:${AWS::AccountId}:*
               - Effect: Allow
                 Action:
                   - iam:PassRole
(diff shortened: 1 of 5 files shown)


apps/metricstream/template.yaml
```diff
@@ -224,6 +226,7 @@ Resources:
                 Resource: !GetAtt 'DeliveryStream.Arn'
   MetricStream:
     Type: AWS::CloudWatch::MetricStream
+    Condition: NotDeployLambda
```
obs-gh-alexlew, line 229: do we need to add a whole variable for this? can't we just do `!DeployLambda`

pkg/handler/metricsconfigurator/handler.go
```diff
@@ -261,6 +270,26 @@ func (h Handler) getDatasource(token *string, observeDomainName string, client *
 	return bodyBytes, nil
 }
 
+func (h Handler) makeMetricGroups(MetricsFilters []types.MetricStreamFilter) [][]types.MetricStreamFilter {
+	// create appropriate number of metric streams for metrics, assign metrics to streams
+
+	currentNameCount := 0
+	filterGroups := make([][]types.MetricStreamFilter, 0)
+	currentFilterGroup := make([]types.MetricStreamFilter, 0)
+	for _, filter := range MetricsFilters {
```
obs-gh-alexlew, line 279: feels like we could do a flatten and then a chunk here no? https://pkg.go.dev/slices@master#Chunk

apps/metricstream/template.yaml
obs-gh-virjramakrishnan: I tried `Condition !Not DeployLambda` and the sam app threw an error - it seems it wants the condition to be a string that refers to an existing condition.

pkg/handler/metricsconfigurator/handler.go
obs-gh-virjramakrishnan: ah, I didn't know this existed, but it's much cleaner, I'll refactor accordingly!

obs-gh-virjramakrishnan: Ah wait, I think that Chunk would not work here - two reasons:
1. We would end up with metrics from the same namespace in different streams. We'd need to make a `types.MetricStreamFilter` for the same namespace in two different streams.
2. AWS counts the namespace itself towards the 1000 (see [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricStreamFilter.html)) so there would be some pretty gnarly cases (adding the namespaces in takes us over the boundary).

obs-gh-alexlew (approved):

pkg/handler/metricsconfigurator/handler.go
obs-gh-alexlew: yeah so then this basically boils down to binpacking, theoretically its possible to be more efficient but we don't really care about how many metric streams it takes us so this simple approach is probably best

PR #391, opened 2025-02-24
feat: update role name for filedrop to include -filedrop suffix

obs-gh-alexlew:


apps/forwarder/template.yaml
obs-gh-ruoyuchen, line 1: I think we don't need to change NameOverride

obs-gh-ruoyuchen: Discussed in slack thread.

obs-gh-ruoyuchen (approved):

integration/tests/forwarder.tftest.hcl
```diff
@@ -127,7 +127,7 @@ run "install_forwarder" {
       SourceTopicArns      = "arn:aws:sns:${run.setup.region}:${run.setup.account_id}:*"
       ContentTypeOverrides = "${var.override_match}=${var.override_content_type}"
       SourceKMSKeyArns     = "${join(",", [for k, v in run.sources.buckets : v.kms_key.arn if v.kms_key != null])}"
-      NameOverride         = run.setup.id
+      NameOverride         = run.setup.short
```
obs-gh-mattcotter, line 130: Where does this come from?

obs-gh-mattcotter (approved):

integration/tests/forwarder.tftest.hcl
obs-gh-alexlew: From an action that we produce, it gets set as one of the outputs of that action

PR #399, opened 2025-05-30
build(deps): Update Go version to 1.23.7 and golang.org/x/net to v0.33.0 - OB-45787 and OB-45788

obs-gh-justindaines:
## Changes
• Updated Go version in `go.mod` to `1.23.7`
• Upgraded `golang.org/x/net` to `v0.33.0`
• Ran `go mod tidy` and `go mod vendor` to sync dependencies
• Updated `DEVELOPER.md` to clarify how to run full and single integration tests using `make test-integration`. `Makefile` was updated in PR https://github.com/observeinc/aws-sam-apps/pull/294, but the `Testing` section in `DEVELOPER.md` was not updated to reflect the changes.

## Testing
• Verified builds pass with Go `1.23.7`
• Ran all integration tests via `make test-integration`
• I manually ran `sam deploy` to validate apps are still working.


obs-gh-mattcotter:
> I agree, I can fix the build. Would it be acceptable to address it in a separate PR? It may take some time to investigate, and this PR resolves a few CVEs that I’d like to address sooner rather than later.

Sorry for the delayed reply, I missed the first notification. We should never leave our automatic checks in a broken state. If you need help investigating and this is truly urgent then please escalate this one.

PR #403, opened 2025-07-07
feat: add support for customMetrics to metricstreamconfigurator

obs-gh-alexlew:


pkg/handler/metricsconfigurator/handler_test.go
```diff
@@ -298,6 +298,43 @@ func TestParseResponse(t *testing.T) {
...
+								"awsServiceMetricsList": [
+									{
+										"namespace": "AWS/EC2",
+										"metricNames": ["CPUUtilization", "NetworkIn"]
+									}
+								],
+								"customMetricsList": [
+									{
+										"namespace": "Custom/Application",
+										"metricNames": ["RequestCount", "ErrorRate"]
```
obs-gh-ruoyuchen, line 318: FE needs to use ["*"] for metricNames because it could NOT know all available custom metricNames in advance.

obs-gh-ruoyuchen: After discussion, FE will use [] to select all the metrics.

obs-gh-ruoyuchen (approved):

PR #411, opened 2025-10-06
add new regions with links to table

obs-gh-virjramakrishnan:
Provide the quick create links for cloudformation stacks in new regions.


obs-gh-justindaines (approved):
LGTM

PR #419, opened 2026-01-28
fix: CloudWatch Log Group subscription handling

obs-gh-justindaines:
## Summary

Improves CloudWatch Log Group subscription lifecycle management in the subscriber Lambda so subscription filters stay aligned with the current include/exclude pattern configuration across stack updates.

Made-with: Cursor

## Changes

If you update a stack and narrow the log group patterns, subscriptions that no longer match will now be automatically deleted. Previously they would have been orphaned.

* Automatic cleanup on pattern narrowing: when stack configuration is updated to narrow `log_group_name_patterns` (or expand excludes), subscriptions that no longer match are now removed automatically instead of being left orphaned.
* Dedicated cleanup path in subscriber: added cleanup request handling to scan log groups and remove stale subscription filters.
* Chunked/continuation cleanup support: cleanup now supports tokenized continuation (scanToken) and bounded work per invocation (maxGroupsPerInvocation), with optional jobId correlation across continuation messages.
* Enhanced update-event handling: stack update flows now trigger subscription reconciliation so adds/removals are applied when patterns change.
* Test coverage improvements: added tests to validate cleanup continuation behavior and no-continuation completion behavior in the subscriber handler.

## Testing

- Verified subscription filters are created for matching log groups.
- Verified subscription filters are updated when include/exclude patterns change.
- Verified subscriptions are removed from groups that no longer match after pattern narrowing.
- Verified cleanup continuation is enqueued when work exceeds per-invocation limit.
- Verified cleanup completes without continuation when all work fits in a single invocation.

---
Testing output:
```
Building codeuri: /aws-sam-apps/bin/linux_arm64 runtime: provided.al2023 metadata: {'BuildMethod': 'makefile'} architecture: arm64 functions: Forwarder/Forwarder
Forwarder: Running CustomMakeBuilder:CopySource
Forwarder: Running CustomMakeBuilder:MakeBuild
Forwarder: Current Artifacts Directory : /Users/justin.daines/repos/aws-sam-apps/apps/stack/.aws-sam/build/Forwarder/Forwarder
cp forwarder /Users/justin.daines/repos/aws-sam-apps/apps/stack/.aws-sam/build/Forwarder/Forwarder/bootstrap
Building codeuri: /Users/justin.daines/repos/aws-sam-apps/bin/linux_arm64 runtime: provided.al2023 metadata: {'BuildMethod': 'makefile'} architecture: arm64 functions: LogWriter/Subscriber
Subscriber: Running CustomMakeBuilder:CopySource
Subscriber: Running CustomMakeBuilder:MakeBuild
Subscriber: Current Artifacts Directory : /Users/justin.daines/repos/aws-sam-apps/apps/stack/.aws-sam/build/LogWriter/Subscriber
cp subscriber /Users/justin.daines/repos/aws-sam-apps/apps/stack/.aws-sam/build/LogWriter/Subscriber/bootstrap
Building codeuri: /Users/justin.daines/repos/aws-sam-apps/bin/linux_arm64 runtime: provided.al2023 metadata: {'BuildMethod': 'makefile'} architecture: arm64 functions: MetricStream/MetricsConfigurator
MetricsConfigurator: Running CustomMakeBuilder:CopySource
MetricsConfigurator: Running CustomMakeBuilder:MakeBuild
MetricsConfigurator: Current Artifacts Directory : /Users/justin.daines/repos/aws-sam-apps/apps/stack/.aws-sam/build/MetricStream/MetricsConfigurator
cp metricsconfigurator /Users/justin.daines/repos/aws-sam-apps/apps/stack/.aws-sam/build/MetricStream/MetricsConfigurator/bootstrap

Build Succeeded

Built Artifacts  : .aws-sam/build
Built Template   : .aws-sam/build/template.yaml

Commands you can use next
=========================
[*] Validate SAM template: sam validate
[*] Invoke Function: sam local invoke
[*] Test Function in the Cloud: sam sync --stack-name {{stack-name}} --watch
[*] Deploy: sam deploy --guided

                Managed S3 bucket: aws-sam-cli-managed-default-samclisourcebucket-ezgtejeau5ut
                A different default S3 bucket can be set in samconfig.toml
                Or by specifying --s3-bucket explicitly.
File with same data already exists at 706154c90665fb59ad3261b83d1b2fe5, skipping upload
File with same data already exists at cb569c6993c2a3673a6545b4e2aef3f8, skipping upload
        Uploading to b9827ca2316ba25481e1d03382ab91ce  15101068 / 15101068  (100.00%)
        Uploading to c53c4dc3aafc12bf7ae69fe02f5e1e8a.template  15145 / 15145  (100.00%)
File with same data already exists at dc236ce78602cbf63d488f3fb977db6f, skipping upload
File with same data already exists at 72b2d82ec9a47b3dd0f2391d730f3215.template, skipping upload
File with same data already exists at 875b2a81a2b31c06ab1a5503e622e4f1, skipping upload
File with same data already exists at 9450d65d7b5e7ac73324d04b64dc6715.template, skipping upload
File with same data already exists at b195b7c066bd2554bdac2c0adcc75222, skipping upload
        Uploading to c09b77581f40a2612310fad044fd73d9  14148976 / 14148976  (100.00%)
        Uploading to e83481a4386c3f24f43bad59112069bb.template  19659 / 19659  (100.00%)
File with same data already exists at 78b46ee819d0e4904a18b827bf5b3d97, skipping upload
        Uploading to dda10bf14638cdfc3fd1cbb1129a8e5d  6418167 / 6418167  (100.00%)
        Uploading to f843b5189c1246a5973e223d0afa7f29.template  11118 / 11118  (100.00%)
File with same data already exists at 0ab6a61b8d788fdaa7375e46568ff0ee, skipping upload
File with same data already exists at 3b9f8953d0129efb21a829036d892d48.template, skipping upload

        Deploying with following values
        ===============================
        Stack name                   : jdd-observe-filedrop
        Region                       : us-east-1
        Confirm changeset            : False
        Disable rollback             : False
        Deployment s3 bucket         : aws-sam-cli-managed-default-samclisourcebucket-ezgtejeau5ut
        Capabilities                 : ["CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND"]
        Parameter overrides          : {"SourceBucketNames": "jdd-config-bucket", "DestinationUri": "s3://156098787674-cn1ci575a1yunaub3j5efffnbxi5ausw2a-s3alias/ds1yLFAGFZvOgh0ogvOJ/", "DataAccessPointArn": "arn:aws:s3:us-west-2:802757454165:accesspoint/156098787674", "NameOverride": "jdd-observe-filedrop", "BucketName": "jdd-config-bucket", "MetricStreamFilterUri": "s3://observeinc/cloudwatchmetrics/filters/recommended.yaml", "LogGroupNamePrefixes": "/aws/ecs", "ExcludeLogGroupNamePatterns": "^/aws/lambda/observeinc,^/aws-glue"}
        Signing Profiles             : {}

Initiating deployment
=====================

        Uploading to 69e074b014352f7fdbeacd48be542a2f.template  18172 / 18172  (100.00%)


Waiting for changeset to be created..

CloudFormation stack changeset
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Operation                                                   LogicalResourceId                                           ResourceType                                                Replacement
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ Add                                                       Bucket                                                      AWS::S3::Bucket                                             N/A
+ Add                                                       Forwarder                                                   AWS::CloudFormation::Stack                                  N/A
+ Add                                                       LogWriter                                                   AWS::CloudFormation::Stack                                  N/A
+ Add                                                       MetricStream                                                AWS::CloudFormation::Stack                                  N/A
+ Add                                                       TopicPolicy                                                 AWS::SNS::TopicPolicy                                       N/A
+ Add                                                       TopicSubscription                                           AWS::SNS::Subscription                                      N/A
+ Add                                                       Topic                                                       AWS::SNS::Topic                                             N/A
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Changeset created successfully. arn:aws:cloudformation:us-east-1:460044344528:changeSet/samcli-deploy1769625752/0f191c63-d1db-4428-82ac-684921a734f5


2026-01-28 13:42:44 - Waiting for stack create/update to complete

CloudFormation events from stack operations (refresh every 5.0 seconds)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ResourceStatus                                              ResourceType                                                LogicalResourceId                                           ResourceStatusReason
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  jdd-observe-filedrop                                        User Initiated
CREATE_IN_PROGRESS                                          AWS::SNS::Topic                                             Topic                                                       -
CREATE_IN_PROGRESS                                          AWS::SNS::Topic                                             Topic                                                       Resource creation Initiated
CREATE_COMPLETE                                             AWS::SNS::Topic                                             Topic                                                       -
CREATE_IN_PROGRESS                                          AWS::SNS::TopicPolicy                                       TopicPolicy                                                 -
CREATE_IN_PROGRESS                                          AWS::SNS::TopicPolicy                                       TopicPolicy                                                 Resource creation Initiated
CREATE_COMPLETE                                             AWS::SNS::TopicPolicy                                       TopicPolicy                                                 -
CREATE_IN_PROGRESS                                          AWS::S3::Bucket                                             Bucket                                                      -
CREATE_IN_PROGRESS                                          AWS::S3::Bucket                                             Bucket                                                      Resource creation Initiated
CREATE_COMPLETE                                             AWS::S3::Bucket                                             Bucket                                                      -
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  Forwarder                                                   -
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  LogWriter                                                   -
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  MetricStream                                                -
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  Forwarder                                                   Resource creation Initiated
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  MetricStream                                                Resource creation Initiated
CREATE_IN_PROGRESS                                          AWS::CloudFormation::Stack                                  LogWriter                                                   Resource creation Initiated
CREATE_COMPLETE                                             AWS::CloudFormation::Stack                                  MetricStream                                                -
CREATE_COMPLETE                                             AWS::CloudFormation::Stack                                  Forwarder                                                   -
CREATE_IN_PROGRESS                                          AWS::SNS::Subscription                                      TopicSubscription                                           -
CREATE_IN_PROGRESS                                          AWS::SNS::Subscription                                      TopicSubscription                                           Resource creation Initiated
CREATE_COMPLETE                                             AWS::SNS::Subscription                                      TopicSubscription                                           -
CREATE_COMPLETE                                             AWS::CloudFormation::Stack                                  LogWriter                                                   -
CREATE_COMPLETE                                             AWS::CloudFormation::Stack                                  jdd-observe-filedrop                                        -
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

CloudFormation outputs from deployed stack
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Outputs
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Key                 LogWriterFirehoseArn
Description         LogWriter Kinesis Firehose Delivery Stream ARN. CloudWatch Log Groups subscribed to this Firehose will have their logs batched and written to S3.
Value               arn:aws:firehose:us-east-1:460044344528:deliverystream/jdd-observe-filedrop-LogWriter

Key                 BucketName
Description         Collection Bucket Name. Objects written to this bucket will be forwarded to your destination. This bucket has a strict expiration policy.
Value               jdd-observe-filedrop-bucket-vtcmgaxdlhw9

Key                 TopicArn
Description         Collection Topic ARN. Events written to this SNS Topic will be forwarded to your destination. S3 Object Created events may trigger an object copy towards destination.
Value               arn:aws:sns:us-east-1:460044344528:jdd-observe-filedrop-Topic-e30s2Ttq3GSv

Key                 LogWriterSubscriberArn
Description         Subscriber Function ARN. This function is responsible for log group discovery, filtering and subscription.
Value               arn:aws:lambda:us-east-1:460044344528:function:jdd-observe-filedrop-LogWriter

Key                 SubscriberLogGroupName
Description         LogWriter Subscriber Log Group Name. This log group contains useful information for debugging the Subscriber function.
Value               /aws/lambda/jdd-observe-filedrop-LogWriter

Key                 ForwarderQueueArn
Description         Forwarder Queue ARN. Events sent to this queue will be forwarded to Observe. s3:ObjectCreated events will initiate an object copy to destination.
Value               arn:aws:sqs:us-east-1:460044344528:jdd-observe-filedrop

Key                 ForwarderLogGroupName
Description         Forwarder Log Group Name. This log group contains useful information for debugging the Forwarder Lambda.
Value               /aws/lambda/jdd-observe-filedrop

Key                 LogWriterSubscriberQueueArn
Description         LogWriter Subscriber Queue ARN. This queue is used by the subscriber function to fan out execution of subscription requests.
Value               arn:aws:sqs:us-east-1:460044344528:jdd-observe-filedrop-LogWriter

Key                 ForwarderArn
Description         Forwarder Function ARN. This is the Lambda function responsible for forwarding objects to Observe.
Value               arn:aws:lambda:us-east-1:460044344528:function:jdd-observe-filedrop

Key                 LogWriterFirehoseLogGroupName
Description         LogWriter Firehose Log Group Name. These logs may contain useful information for debugging Firehose delivery to S3.
Value               /aws/firehose/jdd-observe-filedrop-LogWriter

Key                 ForwarderRoleArn
Description         Forwarder Role ARN. This role will be assumed by the Forwarder Lambda Function in order to write data to destination.
Value               arn:aws:iam::460044344528:role/jdd-observe-filedrop-filedrop

Key                 LogWriterDestinationRoleArn
Description         ARN for IAM Role to be assumed by CloudWatch for log delivery. This value is required when configuring a subscription towards the Firehose Delivery Stream.
Value               arn:aws:iam::460044344528:role/jdd-observe-filedrop-LogWriter-1FVW-DestinationRole-DnmfeCJy2go0
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Successfully created/updated stack - jdd-observe-filedrop in us-east-1
```


pkg/handler/subscriber/cloudformation.go
```diff
@@ -53,12 +58,15 @@ func makeStrSlice(item any) ([]*string, error) {
...
+		// Skip empty strings (CloudFormation passes [""] for empty parameters)
+		if s != "" {
+			ret = append(ret, &s)
+		}
 	}
 	return ret, nil
 }
 
-// HandleCloudFormation triggers discovery on CloudFormation stack updates
+// HandleCloudFormation triggers cleanup and discovery on CloudFormation stack updates, and cleanup on deletes
```
obs-gh-virjramakrishnan: Did you see the note in the logwriter [README](https://github.com/observeinc/aws-sam-apps/blob/main/docs/logwriter.md#deleting-existing-subscriptions-on-uninstall) by any chance? I think that changing `ExcludeLogGroupMatches` to `*` should work for removing log groups on uninstall. Updating is another matter, but for at least one of the customer's concerns (about delete not working to clear subscription filters) we can offer immediate help.

pkg/handler/subscriber/cloudformation.go
```diff
@@ -68,24 +76,194 @@ func (h *Handler) HandleCloudFormation(ctx context.Context, ev *CloudFormationEv
...
-		return h.HandleDiscoveryRequest(ctx, &req)
+
+		logger.Info("discovering log groups",
+			"patterns", req.LogGroupNamePatterns,
+			"prefixes", req.LogGroupNamePrefixes,
+			"excludes", excludePatterns)
+
+		handlerResp, handlerErr = h.HandleDiscoveryRequest(ctx, &req)
+
+	case cfn.RequestDelete:
```
obs-gh-virjramakrishnan: I thought we would do a cleanup on delete? confused here. When the stack is being deleted, would we not get a delete request?

pkg/handler/subscriber/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+		// No subscription from us, nothing to clean up
+		return nil
+	}
+
+	stats.SubscriptionsFound.Add(1)
+
+	// Check if this log group should still be subscribed based on current patterns
+	shouldExist := h.logGroupNameFilter(logGroupName)
+
+	if shouldExist && !deleteAll {
```
obs-gh-virjramakrishnan: this seems very similar to the logic we have in the discovery request processing, in `SubscriptionFilterDiff`: see [here](https://github.com/observeinc/aws-sam-apps/blob/3db18f7ff65a4329560711d8f77a4ac1fe863af7/pkg/handler/subscriber/subscription.go#L61). The key difference between the logic in the cleanup function versus in `SubscriptionFilterDiff` is that in the former, we consider every single log group, regardless of whether they would have been subscribed to before, whereas in `SubscriptionFilterDiff`, we only consider the log groups we now should be subscribed to. I think this adds some duplication of logic which we could avoid.

My suggestion would be: have different kinds of discovery events, maybe via a field in the event, let's say called `fullyPrune`. When a discovery event is triggered via a stack update, set it to true, if it's part of the schedule, set it to false. If the field is true, we can do the exhaustive search we have here, else we do the limited search we do in `SubscriptionFilterDiff`. This is just an idea, so please let me know if you agree!

obs-gh-virjramakrishnan (commented):
Super exciting that this works! I don't think I fully understand all of the code, would be useful to talk through some of it.

pkg/handler/subscriber/cloudformation.go
obs-gh-justindaines: Thank you for pointing that out, I can try that approach if we would rather leave subscriptions as is for now?

I was originally hoping to address the subscriptions for this repo and `terraform-aws-collection`, but I’m now realizing this may be a larger effort than I initially anticipated.

pkg/handler/subscriber/cloudformation.go
```diff
@@ -68,24 +76,194 @@ func (h *Handler) HandleCloudFormation(ctx context.Context, ev *CloudFormationEv
...
+
+	// Return the handler error if there was one
+	if handlerErr != nil {
+		return nil, handlerErr
+	}
+
+	return handlerResp, nil
+}
+
+// buildLogGroupFilter creates a FilterFunc from the given patterns and prefixes.
```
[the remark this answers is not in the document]
obs-gh-justindaines: I've refactored to eliminate the duplication:

Update to us a shared `BuildLogGroupFilter(patterns, prefixes, excludePatterns []string) FilterFunc` function in `config.go`.
Updated `Config.LogGroupFilter()` to delegate to the shared function.

pkg/handler/subscriber/cleanup.go
obs-gh-justindaines: Sorry, I didn't intend to keep that comment. I removed the comment `Kept CleanupRequest for backwards compatibility.`.

pkg/handler/subscriber/cloudformation.go
obs-gh-justindaines: I agree, I think it would make sense to also delete subscriptions here on stack delete. I'll work on an update for this.

pkg/handler/subscriber/discovery.go
```diff
@@ -32,6 +34,19 @@ func (h *Handler) HandleDiscoveryRequest(ctx context.Context, discoveryReq *Disc
 		resp.Discovery.Subscription = new(SubscriptionStats)
 	}
 
+	// If FullyPrune is enabled, scan ALL log groups to find and remove stale subscriptions
+	// This is used during stack updates when patterns change
+	if discoveryReq.FullyPrune {
+		logger.Info("fully pruning stale subscriptions")
+		resp.Discovery.Cleanup = new(CleanupStats)
```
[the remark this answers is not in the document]
obs-gh-justindaines: I do think It makes more sense to update to use `subscriptionStats`. I'll work on updating.

pkg/handler/subscriber/cloudformation.go
obs-gh-justindaines: I made the updates, the CloudFormation handler now supports cleanup on `RequestDelete`. The challenge was distinguishing a true stack delete from a resource replacement during `UPDATE` (which also sends a `DELETE` for the old resource). The solution uses `lambdacontext.LogStreamName`: during replacement, the DELETE event's `PhysicalResourceID` carries the OLD log stream name, but the Lambda is running in a NEW log stream. When they match, it's a true delete and we run `HandleCleanupRequest`(DeleteAll: true). When they differ, we skip cleanup.

For Terraform deployments, cleanup is handled separately by `terraform-aws-collection` PR [#224](https://github.com/observeinc/terraform-aws-collection/pull/224) via a `local-exec` provisioner in `cleanup.tf` that sends a cleanup message to SQS on terraform destroy.

pkg/handler/subscriber/subscription.go
```diff
@@ -44,9 +51,28 @@ func (h *Handler) SubscribeLogGroup(ctx context.Context, logGroup *LogGroup, sta
...
 	stats.Processed.Add(1)
 
-	output, err := h.Client.DescribeSubscriptionFilters(ctx, &cloudwatchlogs.DescribeSubscriptionFiltersInput{
-		LogGroupName: &logGroup.LogGroupName,
-	})
+	var (
+		output *cloudwatchlogs.DescribeSubscriptionFiltersOutput
+		err    error
+	)
+	for attempt := 1; attempt <= cloudWatchAPIMaxAttempts; attempt++ {
```
obs-gh-virjramakrishnan: This code is being repeated a bit, could we make this a helper function?

Also, according to my AI: "The AWS SDK v2 already retries throttling errors. By default, the SDK's retry.Standard retryer handles ThrottlingException, TooManyRequestsException, etc. with exponential backoff." Is there a specific reason for this extra retrying logic?

pkg/handler/subscriber/handler.go
```diff
@@ -103,5 +109,93 @@ func New(cfg *Config) (*Handler, error) {
...
+func (h *Handler) callCloudWatch(ctx context.Context, fn func() error) error {
+	if h.limiter != nil {
+		if err := h.limiter.Wait(ctx); err != nil {
+			return err
+		}
+	}
+	return fn()
+}
+
+type apiLimiter struct {
```
obs-gh-virjramakrishnan: I think we might be able to get away with using Go's own: `golang.org/x/time/rate` here - might handle edge cases better.

pkg/handler/subscriber/retry.go
```diff
@@ -0,0 +1,60 @@
+package subscriber
```
obs-gh-virjramakrishnan, line 1: this is super cool functionality, just wondering what the need is for custom retry logic? Did we receive any comments from the customer alluding to unreliable retries? I don't want to add a big surface area for us to maintain here.

pkg/handler/subscriber/request.go
```diff
@@ -120,3 +141,18 @@ func (d *DiscoveryRequest) ToDescribeLogInputs() (inputs []*cloudwatchlogs.Descr
 
 	return inputs
 }
+
+// CleanupRequest scans all log groups and removes subscriptions that no longer match the configured patterns.
+type CleanupRequest struct {
+	// DryRun if true, will only log what would be deleted without actually deleting
+	DryRun bool `json:"dryRun,omitempty"`
+	// DeleteAll if true, will delete all subscriptions regardless of whether they match patterns
+	DeleteAll bool `json:"deleteAll,omitempty"`
```
obs-gh-virjramakrishnan: I think we never set DeleteAll to false - and if we do, it's essentially the same thing as fullyPrune (I think). Do we need this option?

obs-gh-virjramakrishnan (commented):
functionally this looks fine, just some comments on the code structure.

pkg/handler/subscriber/handler.go
obs-gh-justindaines: Thank you. I switched to `golang.org/x/time/rate`. The custom `apiLimiter` (~60 lines) is removed; `rate.NewLimiter` is a drop-in replacement and the Wait(ctx) API is identical.

pkg/handler/subscriber/retry.go
obs-gh-justindaines: The retry logic is specifically needed for `OperationAbortedException` (concurrent modification conflicts within CWL itself), which the SDK doesn't handle. The repeated loops into a single `callCloudWatchWithRetry` helper to reduce the maintenance surface. 

I updated the retry policy in one place.

I did have earlier failures during testing without the retry logic.

pkg/handler/subscriber/subscription.go
obs-gh-justindaines: Great call, I updated to consolidate. The 5 retry loops across `subscription.go` and `cleanup.go` are now all consolidated into a `callCloudWatchWithRetry` helper in `retry.go`.

CoCo response: On the SDK retry question: the main driver for keeping application-level retries is `OperationAbortedException`. CloudWatch Logs returns this when two workers concurrently try to modify subscriptions for the same log group (e.g. during discovery with multiple workers). The SDK v2 `retry.Standard` does not cover this error code — only throttling/transient ones. The throttling codes are kept in the list as defense-in-depth, since our retries also re-enter the rate limiter, which the SDK's internal retry bypasses.

pkg/handler/subscriber/request.go
obs-gh-justindaines: `DeleteAll` is semantically different from `FullyPrune` (on `DiscoveryRequest`). FullyPrune option will "scan all log groups, but still only remove subscriptions that no longer match patterns." `DeleteAll` will "remove our subscription regardless of whether the log group still matches". 

`DeleteAll` is used during CloudFormation stack deletion (cloudformation.go:163) to guarantee all subscriptions are removed on teardown, even for log groups that would still match the configured patterns. Dropping it would leave stale subscriptions behind when a stack is deleted.

obs-gh-virjramakrishnan: When DeleteAll is set to true, it is different, yes. Here's how I see it:
1. Cleanup, DeleteAll true: delete all log groups
2. Cleanup, DeleteAll false: delete all log groups that do not match filters
3. DiscoveryRequest, FullPrune true: delete all log groups that do not match filters, AND subscribe to new ones that may match
4. DiscoveryRequest, FullPrune false: subscribe to new groups that now match filters

1 is necessary, as your comment points out. 3 is necessary for when there's a stack update and we'd like for things to stay consistent. 4 is sufficient for when there is no update but the user may have simply added more log groups that should now be subscribed to. I can't think of a use case for 2 though, where we wouldn't want to use 3 That code path is also unused (we always call cleanup with deleteAll true). Therefore, my suggestion is that we make the DeleteAll behavior the default and remove the variable. What say you?

pkg/handler/subscriber/retry.go
obs-gh-virjramakrishnan: gotcha, makes sense!

pkg/handler/subscriber/request.go
obs-gh-justindaines: Doh, you're right. I updated the code to remove the `DeleteAll` parameter.

**Changes made:**
- Removed `DeleteAll` field from `CleanupRequest` 
- Simplified `CleanupLogGroup()` to always delete all subscriptions
- Updated Terraform cleanup payload to remove `"deleteAll":true`
- Updated tests and documentation

**New behavior:**
- **Cleanup** (during destroy): Always removes ALL subscriptions managed by this handler
- **Discovery with `fullyPrune: true`** (during apply/update): Scans all groups, removes non-matching subscriptions, adds new ones
- **Discovery with `fullyPrune: false`**: Only adds new subscriptions

This matches the actual usage pattern where we always called cleanup with `deleteAll=true`. The API is now clearer - if you want to clean up, it's destructive. If you want smart reconciliation, use discovery with fullyPrune.

Use case #2 (DeleteAll=false) as you noted was never used in practice, so removing it simplifies the codebase without losing functionality.

All tests pass ✅

obs-gh-justindaines (commented):
Updated to remove the `DeleteAll` parameter.

- Removed `DeleteAll` field from `CleanupRequest` 
- Simplified `CleanupLogGroup()` to always delete all subscriptions
- Updated Terraform cleanup payload to remove `"deleteAll":true`
- Updated tests and documentation

**New behavior:**
- **Cleanup** (during destroy): Always removes ALL subscriptions managed by this handler
- **Discovery with `fullyPrune: true`** (during apply/update): Scans all groups, removes non-matching subscriptions, adds new ones
- **Discovery with `fullyPrune: false`**: Only adds new subscriptions

This matches the actual usage pattern where we always called cleanup with `deleteAll=true`. The API is now clearer - if you want to clean up, it's destructive. If you want smart reconciliation, use discovery with fullyPrune.

Use case #2 (DeleteAll=false) as you noted was never used in practice, so removing it simplifies the codebase without losing functionality.

All tests pass ✅

obs-gh-virjramakrishnan (approved):
LGTM! ship it

PR #445, opened 2026-04-08
build: add SKIP_REGIONS to allow skipping unavailable regions during …

obs-gh-virjramakrishnan:
…release

Adds a SKIP_REGIONS variable (defaulting to empty) that filters regions out of AWS_REGIONS via make's filter-out. Exposed as a workflow_dispatch input on the Release workflow so broken regions can be skipped without code changes.

Made-with: Cursor

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,6 +11,10 @@ on:
         type: boolean
         description: 'Skip tests.'
         default: false
+      skip_regions:
+        type: string
+        description: 'Space-separated list of regions to skip'
+        default: ''
  
 jobs:
   version:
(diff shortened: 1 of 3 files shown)


obs-gh-mattcotter (approved):
What is the follow up if we skip a region? I wonder if we also need the compliment of this; a way to release to only select regions.

obs-gh-virjramakrishnan:
I think that the follow up is to release to those certain regions. A couple of cases:
1. We add a new region
  - If there's a new version to be released, no problem, just run the release
  - If not, we should have a way to add the latest release to just this new region (stand up a release in just this region, then point the `latest` alias to it
2. We want to exclude a region
  - We can use the feature in this PR
3. We want to release to a specific, existing region (that doesn't have latest for some reason)
  - Using the same feature as in 1., we should be able to release the existing latest release for just that region.
  
  I'll have a look at the makefiles and figure out the best way to do that.

PR #450, opened 2026-04-12
feat: Update Tag release to allow SKIP_REGIONS

obs-gh-justindaines:
## Problem

The release workflow was failing during the publish step when attempting to pull SAM manifests from the `me-south-1` region, which is currently experiencing availability issues.

The workflow's `upload` step correctly respected the `skip_regions` input parameter, but the `publish` step (which runs `make tag`) did not pass the `SKIP_REGIONS` environment variable, causing it to attempt operations on all regions including unavailable ones.

Note: Fix updated using Augment

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -90,7 +90,7 @@ jobs:
     - name: AWS Info
       run: aws sts get-caller-identity
 
-    - name: Tag release 
+    - name: Tag release
       id: build
       run: |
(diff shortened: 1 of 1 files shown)


obs-gh-brett-blakeney:
Approved

obs-gh-brett-blakeney (commented):
Looks good

obs-gh-brett-blakeney (approved):

PR #452, opened 2026-04-13
fix(forwarder): encode '+' as '%2B' in CopySource for x-amz-copy-source

obs-gh-virjramakrishnan:
S3 treats unencoded '+' in copy-source headers as a space, causing NoSuchKey for keys containing '+'. Decode back to '+' for policy filters, override regex matching, and s3http GetObject key.

Made-with: Cursor

### DIFF
--- a/pkg/handler/forwarder/handler.go
+++ b/pkg/handler/forwarder/handler.go
@@ -50,6 +50,17 @@ type Handler struct {
 	MaxConcurrentTasks int
 }
 
+// encodeCopySourceKey URL-encodes each path segment of an S3 key for use in
+// the x-amz-copy-source header, preserving '/' separators between segments.
+func encodeCopySourceKey(rawKey string) string {
+	rawKey = strings.TrimPrefix(rawKey, "/")
+	segs := strings.Split(rawKey, "/")
+	for i := range segs {
+		segs[i] = url.QueryEscape(segs[i])
+	}
+	return strings.Join(segs, "/")
+}
+
 // GetCopyObjectInput constructs the input struct for CopyObject.
 func GetCopyObjectInput(source, destination *url.URL) *s3.CopyObjectInput {
 	if source == nil || destination == nil {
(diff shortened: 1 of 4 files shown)


pkg/handler/forwarder/override/rule.go
```diff
@@ -29,7 +29,7 @@ type Filter struct {
 
 // Match input object.
 func (f *Filter) Match(input *s3.CopyObjectInput) bool {
-	if f.Source != nil && !f.Source.MatchString(aws.ToString(input.CopySource)) {
+	if f.Source != nil && !f.Source.MatchString(strings.ReplaceAll(aws.ToString(input.CopySource), "%2B", "+")) {
```
obs-gh-mattcotter: Should we do a URL encoding instead of this one replace? Either way, can this live in a helper method in case we need it somewhere else?

obs-gh-virjramakrishnan: I think we should avoid doing a full encoding, because the slashes also get encoded. To avoid the entire thing getting encoded, I have now split on slashes, and made it a helper function! I also checked and confirmed that the + character is the bug by manually doing a copy operation, and this avoids that.

pkg/handler/forwarder/handler.go
```diff
@@ -147,7 +158,7 @@ func (h *Handler) ProcessRecord(ctx context.Context, record *events.SQSMessage)
 
 		copyInput := GetCopyObjectInput(sourceURL, h.DestinationURI)
 
-		if !h.ObjectPolicy.Allow(aws.ToString(copyInput.CopySource)) {
+		if !h.ObjectPolicy.Allow(sourceURL.Host + sourceURL.Path) {
```
obs-gh-mattcotter, line 161: Looking at this change, does the Object Policy expect to be matching on (eg) `2026-04-08T20:50:00+00:00`? Or on `2026-04-08T20%3A50%3A00%2B00%3A00`?

Will the sourceURL be decoded here?

obs-gh-virjramakrishnan: the object policy expects unencoded strings (in fact it does a regex match under the hood). `sourceURL.Host + sourceURL.Path` is unencoded, we preserve this because we look at the original variable sourceURL, rather than the `CopySource` which now has the 'section-wise' encoding that was called for.

obs-gh-mattcotter (approved):

PR #456, opened 2026-04-23
feat: add pollerconfigurator Lambda for managing Observe pollers

obs-gh-virjramakrishnan:
## Summary

- Adds a new `pollerconfigurator` Lambda function that manages CloudWatch Metrics poller lifecycle in the Observe backend via the GQL API.
- Used as a CloudFormation custom resource to create, update, and delete pollers when stacks are deployed across an organization.
- Includes comprehensive unit tests for GQL mutation building, handler logic, S3 config download, and CFN response reporting.

This is the first in a sequence of PRs extracted from #444 (add-stacksets). Each PR in the chain is independently releasable.

## What's included

- `cmd/pollerconfigurator/main.go` — Lambda entry point
- `pkg/handler/pollerconfigurator/` — handler, GQL client, config types, S3 config downloader, CFN request/response types
- `pkg/lambda/pollerconfigurator/lambda.go` — Lambda wrapper
- `variables.mk` / `Dockerfile.all-binaries` — build pipeline additions (additive only)

## What's NOT included

- No new SAM app template — the binary gets built but is not yet referenced by any CloudFormation template
- `.gitignore` changes deferred to a later PR in the chain

## Test plan

- [x] Cherry-picks cleanly onto current `main`
- [x] `go build ./...` succeeds
- [x] `go vet` passes
- [x] All 19 new tests pass (`go test ./pkg/handler/pollerconfigurator/...`)
- [x] All existing tests still pass (`go test ./...`)
- [ ] CI passes (lint, test, sam-validate, upload, integration)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/handler/pollerconfigurator/gql.go
+++ b/pkg/handler/pollerconfigurator/gql.go
@@ -1,0 +1,259 @@
+package pollerconfigurator
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/go-logr/logr"
+)
+
+const bearerTokenFormat = "Bearer %s %s"
+
+type graphQLRequest struct {
+	Query string `json:"query"`
+}
+
+type createPollerResponse struct {
+	Data struct {
+		CreatePoller struct {
+			ID   string `json:"id"`
+			Name string `json:"name"`
+		} `json:"createPoller"`
+	} `json:"data"`
+	Errors []struct {
+		Message string `json:"message"`
+	} `json:"errors"`
+}
+
+type updatePollerResponse struct {
+	Data struct {
+		UpdatePoller struct {
+			ID   string `json:"id"`
+			Name string `json:"name"`
+		} `json:"updatePoller"`
+	} `json:"data"`
+	Errors []struct {
+		Message string `json:"message"`
+	} `json:"errors"`
+}
+
+type deletePollerResponse struct {
+	Data struct {
+		DeletePoller struct {
+			Success bool `json:"success"`
+		} `json:"deletePoller"`
+	} `json:"data"`
+	Errors []struct {
+		Message string `json:"message"`
+	} `json:"errors"`
+}
+
+func buildQueriesGQL(queries []QueryConfig) string {
+	var parts []string
+	for _, q := range queries {
+		var fields []string
+		fields = append(fields, fmt.Sprintf(`namespace: %q`, q.Namespace))
+
+		if len(q.MetricNames) > 0 {
+			names := make([]string, len(q.MetricNames))
+			for i, n := range q.MetricNames {
+				names[i] = fmt.Sprintf("%q", n)
+			}
+			fields = append(fields, fmt.Sprintf(`metricNames: [%s]`, strings.Join(names, ", ")))
+		}
+
+		if len(q.Dimensions) > 0 {
+			var dims []string
+			for _, d := range q.Dimensions {
+				if d.Value != "" {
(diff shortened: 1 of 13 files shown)


pkg/handler/pollerconfigurator/s3config.go
```diff
@@ -0,0 +1,113 @@
...
+	Name         string        `json:"name"`
+	DatastreamId string        `json:"datastreamId"`
+	Period       int64         `json:"period"`
+	Delay        int64         `json:"delay"`
+	Interval     string        `json:"interval"`
+	Retries      *int64        `json:"retries,omitempty"`
+	Queries      []QueryConfig `json:"queries"`
+}
+
+func s3URIToHTTPS(uri string) (string, error) {
```
obs-gh-mattcotter: Can we use `s3.GetObject` instead here?

pkg/handler/pollerconfigurator/handler.go
```diff
@@ -0,0 +1,254 @@
...
+			if reportErr := h.reportStatus(*req, true, "deleted (skipped poller cleanup: could not retrieve token)"); reportErr != nil {
+				return nil, fmt.Errorf("failed to report success: %w", reportErr)
+			}
+			return []byte{}, nil
+		}
+		return nil, h.reportAndError("failed to retrieve GQL token", req, err)
+	}
+
+	gql := &gqlClient{
+		httpClient:        &http.Client{},
```
obs-gh-mattcotter: I think the default here doesn't have any timeouts. Can you add one so calls to this don't hang?

pkg/handler/pollerconfigurator/gql.go
```diff
@@ -0,0 +1,259 @@
...
+	Errors []struct {
+		Message string `json:"message"`
+	} `json:"errors"`
+}
+
+func buildQueriesGQL(queries []QueryConfig) string {
+	var parts []string
+	for _, q := range queries {
+		var fields []string
+		fields = append(fields, fmt.Sprintf(`namespace: %q`, q.Namespace))
```
obs-gh-mattcotter: I think this would be safer if it used GQL variables: https://graphql.org/learn/queries/#variables
This would cut down on stringification for the queries / mutations (I think they could be hardcoded strings?) and would make the field serialization safer since I believe variables use standard json encoding.

obs-gh-mattcotter (commented):
Overall I think it looks good!

I left a few comments and separately shared my concerns about the GQL. Can you also update the README to reflect your new changes?

pkg/handler/pollerconfigurator/gql.go
obs-gh-virjramakrishnan: got it - I'll update!

pkg/handler/pollerconfigurator/handler_test.go
```diff
@@ -0,0 +1,420 @@
...
+		t.Fatalf("variables missing key %q", key)
+	}
+	m, ok := v.(map[string]interface{})
+	if !ok {
+		t.Fatalf("variables[%q] is %T, want map", key, v)
+	}
+	return m
+}
+
+func assertMapField(t *testing.T, m map[string]interface{}, key, want string) {
```
obs-gh-mattcotter, line 398: It feels like this and `assertVariable` could be combined if `want` was `any` instead of string?

obs-gh-mattcotter (approved):
One last comment, otherwise lgtm!

pkg/handler/pollerconfigurator/handler_test.go
obs-gh-virjramakrishnan: right you are! Will clean that up.

PR #457, opened 2026-04-27
feat: add new metric tag lambda to be used as a firehose transformation for our metricstream app

obs-gh-mattcotter:
This new lambda uses Prometheus YACE to map metrics to resources. The high level flow is:
1. look up all relevant resources in a read-through cache by account / region / namespace
2. use the YACE associator to map metric dimensions to a single resource
3. copy the resource tags onto the metric and emit the resulting json


DEVELOPER.md
```diff
@@ -239,6 +239,32 @@ aws s3api put-bucket-notification-configuration --bucket $SOURCE_BUCKET \
 
 This manual setup mirrors the automated testing environment defined in `forwarder.tftest.hcl`, allowing developers to validate the entire event flow end-to-end.
 
+### Deploying the Full Collection Stack
```
obs-gh-virjramakrishnan, line 242: thanks for adding this!

apps/metricstream/template.yaml
```diff
@@ -95,10 +95,20 @@ Parameters:
     Description: >-
       The token used to retrieve metric configuration.
     Default: ''
+  TransformFunctionArn:
```
obs-gh-virjramakrishnan: why have a lambda ARN here? if we were to define the lambda in metric stream app, I believe we would not need to have this field, we could just reference the lambda directly?

A wider question I guess it makes sense to touch on here: why make a separate app? I feel like this is an addendum to the metricstream app, and therefore doesn't need to be a separate app. We couldn't deploy this app on its own, if I am understanding correctly.

apps/stack/template.yaml
```diff
@@ -234,6 +237,30 @@ Parameters:
...
+  EnableMetricTag:
+    Type: String
+    Description: >-
+      Enable the metrictag Lambda as a Firehose data transformation that
+      enriches CloudWatch metric stream records with AWS resource tags.
+    AllowedValues:
+      - 'true'
+      - 'false'
+    Default: 'false'
+  MetricTagAssumeRoleArn:
```
obs-gh-virjramakrishnan: I think `MetricTagAssumeRoleArn` on its own cannot help in cross-account scenarios. The reason is that it is a single value, meaning that, if it were to be used cross-account, the role it assumes could only possibly allow it to check tags for one account. The user would have to create a separate stackset with all of the roles, then format the stackset so that this field referenced them. I think the correct solution here is just the lambda role you already have in the metrictag stack.

obs-gh-virjramakrishnan (commented):
Looks good, some structural questions. I thought that your metric tagging lambda would be per-metricstream stack, and therefore, to support multi-account, we would need to have the metric tagging lambda deployed in every account/region in the metricstream stack, and the lambdas would then send data on to the central bucket, as before. My understanding is that this doesn't hurt performance (as there could not possibly be any cache hits across accounts, the resources would be disjoint).

In your testing, how how much extra latency did adding these lambdas add?

apps/metricstream/template.yaml
obs-gh-mattcotter: Great call out! I eliminated the second app and then rebased on top of your latest CloudFormation conversion change. Originally I thought this would help with updating but it doesn't play well with our add data portal.

apps/stack/template.yaml
obs-gh-mattcotter: Agreed and removed! Thanks

obs-gh-mattcotter:
> Looks good, some structural questions. I thought that your metric tagging lambda would be per-metricstream stack, and therefore, to support multi-account, we would need to have the metric tagging lambda deployed in every account/region in the metricstream stack, and the lambdas would then send data on to the central bucket, as before. My understanding is that this doesn't hurt performance (as there could not possibly be any cache hits across accounts, the resources would be disjoint).
> 
> In your testing, how how much extra latency did adding these lambdas add?

Yes, this will be per metricstream stack! Latency averaged just under 2s in my testing.

apps/stack/template.yaml
```diff
@@ -48,6 +48,9 @@ Metadata:
           - DatasourceID
           - GQLToken
           - UpdateTimestamp
+          - EnableMetricTag
+          - MetricTagResourceCacheTTLSeconds
+          - MetricTagLambdaS3Key
```
obs-gh-virjramakrishnan: I think the variable `MetricTagLambdaS3Key` is added in two places (I think it makes more sense in the Lambda Code section).

apps/metricstream/template.yaml
```diff
@@ -305,6 +352,7 @@ Resources:
       Runtime: provided.al2023
       Architectures:
         - arm64
+      Timeout: 60
```
obs-gh-virjramakrishnan: Thank you for adding a timeout here. Could you add a timeout to the MetricTag Lambda as well?

pkg/handler/metrictag/enrich/enrich.go
```diff
@@ -0,0 +1,245 @@
...
+
+// New builds an Enricher from config. A nil cfg uses the same defaults as the former env-only bootstrap.
+func New(cfg *Config, logger *slog.Logger) (*Enricher, error) {
+	if logger == nil {
+		logger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
+	}
+	if cfg == nil {
+		cfg = &Config{}
+	}
+	ttl := 10 * time.Minute
```
obs-gh-virjramakrishnan: If ResourceCacheTTLSeconds is set to 0, then the ttl gets set to 10 minutes it seems. However, the comment on the `MetricTagResourceCacheTTLSeconds` variable says that it should disable caching

apps/metricstream/template.yaml
```diff
@@ -325,6 +373,94 @@ Resources:
...
+      Handler: bootstrap
+      Runtime: provided.al2023
+      Architectures:
+        - arm64
+      LoggingConfig:
+        LogGroup: !Ref MetrictagLogGroup
+      Environment:
+        Variables:
+          RESOURCE_CACHE_TTL_SECONDS: !Ref MetricTagResourceCacheTTLSeconds
+          VERBOSITY: '6'
```
obs-gh-virjramakrishnan: Verbosity is hardcoded at the moment, could we make this a lower number, or a stack variable? The reason I ask for this lambda specifically is that (unlike the configurators which run only once per deploy), this lambda will continue to be executed. I'm ok with the configurators having a high verbosity, but I think this lambda should default lower, like the logwriter and forwarder (they set it to 1). They have it wired up as a variable in their own individual stacks, but they do not get it from the parent `stack` stack.

apps/metricstream/template.yaml
```diff
@@ -325,6 +373,94 @@ Resources:
...
+  MetrictagFirehosePermission:
+    Type: AWS::Lambda::Permission
+    Condition: EnableMetricTag
+    Properties:
+      FunctionName: !GetAtt Metrictag.Arn
+      Action: lambda:InvokeFunction
+      Principal: firehose.amazonaws.com
+      SourceAccount: !Ref AWS::AccountId
```
obs-gh-virjramakrishnan: nit: can you scope this down to just the firehose itself with `SourceArn: !GetAtt DeliveryStream.Arn`? Right now, technically any firehose in this account could use this lambda. Not a huge deal, but just for the sake of granting minimal permissions.

obs-gh-virjramakrishnan (commented):
Just a few small comments, looking great otherwise!

obs-gh-virjramakrishnan (approved):
LGTM! 🚢

PR #458, opened 2026-05-05
feat: convert logwriter/metricstream/externalrole to plain CloudFormation

obs-gh-virjramakrishnan:
## Summary

- Converts logwriter, metricstream, and externalrole templates from `AWS::Serverless::Function` to `AWS::Lambda::Function` with S3-based code references. Required because CloudFormation StackSets do not support the SAM transform.
- Adds Lambda ZIP build, S3 upload, and `embed-lambda-defaults.py` post-processing to the packaging pipeline so released templates are self-contained.
- Categorizes apps into `SAM_APPS`, `CFN_APPS`, and `PACKAGEABLE_APPS`; `sam validate` runs on SAM apps while `cfn-lint` validates the new plain CloudFormation templates.
- Updates CI to install `cfn-lint`/`yamllint` and adds the `s3:GetObject` permission needed by the metricstream integration test.

This PR is stacked on top of #462 (FilterUri-via-Lambda). It contains only the plain-CFN packaging work; the metricsconfigurator FilterUri changes have been moved to #462. Once #462 merges, this PR auto-rebases onto main and is independently releasable.

## What's included

**Template changes** (4 files):
- `apps/logwriter/template.yaml` — subscriber Lambda converted to `AWS::Lambda::Function`
- `apps/metricstream/template.yaml` — metricsconfigurator Lambda converted; `LambdaS3BucketPrefix`/`LambdaS3Key` parameters; `HasLambdaCode` / `DeployLambda` conditions
- `apps/externalrole/template.yaml` — pollerconfigurator Lambda converted, cfn-lint W1030 suppressed
- `apps/stack/template.yaml` — passes through `LambdaS3BucketPrefix`/`LambdaS3Key*` to nested stacks

**Build system** (3 files):
- `Makefile` — Lambda ZIP targets, S3 upload during `sam-package`, cfn-lint validation targets
- `variables.mk` — `SAM_APPS`/`CFN_APPS`/`PACKAGEABLE_APPS` split, `LAMBDA_ZIP_BINS`, bucket auto-creation
- `scripts/embed-lambda-defaults.py` — embeds S3 bucket/key defaults into packaged templates

**CI** (2 files):
- `.github/workflows/tests.yaml` — install cfn-lint and yamllint
- `.github/workflows/integration.yaml` — minor adjustments
- `integration/tests/metricstream.tftest.hcl` — add `s3:GetObject` for the Lambda code ZIP

**Docs** (4 files):
- Updated parameter tables for logwriter, metricstream, externalrole, stack

## Deployment notes

- **No breaking changes for existing deployments.** `LambdaS3BucketPrefix` and `LambdaS3Key` default to empty; the release pipeline (`embed-lambda-defaults.py`) populates them automatically.
- Existing single-account `stack` deployments continue to work unchanged.
- The SAM transform is removed from logwriter/metricstream/externalrole only; `config`, `configsubscription`, `forwarder`, and `stack` remain SAM-based.

## Test plan

- [ ] CI passes: `go-test`, `go-lint`, `sam-validate` (now includes `cfn-lint` for CFN apps)
- [ ] `make sam-package-logwriter`, `make sam-package-metricstream`, `make sam-package-externalrole` produce valid templates with embedded Lambda S3 defaults
- [ ] Integration test for metricstream passes with the new `s3:GetObject` permission
- [ ] Existing `stack` deployment upgrades cleanly (nested stacks get new Lambda params with defaults)

Made with [Cursor](https://cursor.com)


pkg/handler/metricsconfigurator/handler.go
```diff
@@ -153,18 +167,102 @@ func (h Handler) Invoke(ctx context.Context, payload []byte) ([]byte, error) {
...
+		_, err := cwClient.DeleteMetricStream(ctx, &cloudwatch.DeleteMetricStreamInput{
+			Name: &name,
+		})
+		if err != nil {
+			var rnf *types.ResourceNotFoundException
+			if errors.As(err, &rnf) {
+				break
+			}
+			logger.Error(err, "failed to delete metric stream, continuing", "name", name)
+			break
```
obs-gh-mattcotter: Is this a typo? I wonder if this should be

```suggestion
			continue
```

pkg/handler/metricsconfigurator/filteruri.go
```diff
@@ -0,0 +1,99 @@
...
+	path := strings.TrimPrefix(uri, "s3://")
+	parts := strings.SplitN(path, "/", 2)
+	if len(parts) < 2 || parts[1] == "" {
+		return "", fmt.Errorf("invalid S3 URI: must contain bucket and key")
+	}
+	return fmt.Sprintf("https://%s.s3.amazonaws.com/%s", parts[0], parts[1]), nil
+}
+
+func downloadFilterYAML(ctx context.Context, filterURI string) ([]byte, error) {
+	url, err := s3URIToHTTPS(filterURI)
```
obs-gh-mattcotter, line 38: Can this use s3 getObject? I recall you making a similar change somewhere else

apps/metricstream/template.yaml
```diff
@@ -230,30 +246,6 @@ Resources:
                   - firehose:PutRecord
                   - firehose:PutRecordBatch
                 Resource: !GetAtt 'DeliveryStream.Arn'
-  MetricStream:
```
obs-gh-mattcotter: Is this MetricStream resource gone entirely?

pkg/handler/metricsconfigurator/handler.go
obs-gh-virjramakrishnan: I actually realized a larger problem which had been lurking in this delete the whole time. For the first PR ([here](https://github.com/observeinc/aws-sam-apps/pull/462)), I am going to disable the delete metric stream. This was anyway not a thing before - metric streams created by the stack would be orphaned on stack delete. I'll merge in the first PR with no delete logic, fast follow with correct logic for delete, and then this PR should have nothing to do with that.

It turns out the metric stream delete was making some assumptions about how the delete works. I'll go into more detail in the other PR.

apps/metricstream/template.yaml
obs-gh-virjramakrishnan: Yes. Will go into more detail in the other [PR](https://github.com/observeinc/aws-sam-apps/pull/462), but since the lambda now creates everything, regardless of if the user provides an s3 file or not, there's no need to specify the resource here. This does add some issues when it comes to deleting, which I will address in a separate PR.

pkg/handler/metricsconfigurator/filteruri.go
obs-gh-virjramakrishnan: good point. I will change this in the PR where I switch over s3 processing to lambda since we decided to split it.

obs-gh-virjramakrishnan:
Closed in favor of the split: the FilterUri-via-Lambda half landed as #462 and the SAM → plain CloudFormation conversion half is now #464. This branch's history is preserved on `viraj/convert-to-plain-cfn` if anyone wants to reference it.

PR #460, opened 2026-05-19
fix: ALB logs fail to decode when HTTPS destination due to missing gzip content-encoding

### LINKED ISSUE #459: ALB logs fail with "invalid character 'x' in string escape code" when using HTTPS destination
## Description

When configuring the forwarder to send AWS Application Load Balancer (ALB) access logs to an HTTPS `DESTINATION_URI`, the Lambda fails to process the logs, throwing an `invalid character 'x' in string escape code` error.

This issue is completely silent when `DESTINATION_URI` is an S3 URI (`s3://`) because the payload bypasses the decode pipeline and raw bytes are copied directly. The bug only surfaces when the full decode pipeline runs for HTTPS destinations.

## Steps to Reproduce

1. Deploy the forwarder Lambda with an HTTPS `DESTINATION_URI` (e.g., `https://<endpoint>`).
2. Configure an AWS Application Load Balancer to write access logs to an S3 bucket monitored by the forwarder.
3. Generate traffic so ALB access logs are written to S3. Note that ALB log filenames contain dots in the load balancer name segment, e.g.:
   ```
   <account-id>_elasticloadbalancing_<region>_app.<alb-name>.<hexid>_<timestamp>_<ip>_<random>.log.gz
   ```
4. Observe the Lambda execution logs.

## Expected Behavior

The forwarder identifies the file as a gzipped ALB log, applies the correct preset (`application/x-aws-elasticloadbalancing` + `gzip`), decompresses it, and forwards the contents to the HTTPS destination successfully.

The uncompressed content of an ALB `.log.gz` is space-separated plain text, one record per line, e.g.:

```
https 2024-03-15T10:43:05.881204Z app/my-app-alb/1a2b3c4d5e6f7a8b 198.51.100.72:44812 10.0.2.101:8443 0.002 0.045 0.001 200 200 638 24310 "POST https://app.example.com:443/api/v2/users HTTP/1.1" "Mozilla/5.0 (Windows NT 10.0; Win64; x64)" ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2 arn:aws:elasticloadbalancing:<region>:<account-id>:targetgroup/my-app-tg/aabbccddeeff0011 "Root=1-65f4a311-7c2e4b1a9f3d5e8b0c6a1f2e" "app.example.com" "arn:aws:acm:<region>:<account-id>:certificate/aaaabbbb-cccc-dddd-eeee-ffffaaaabbbb" 1 2024-03-15T10:43:05.834000Z "forward" "-" "-" "10.0.2.101:8443" "200" "-" "-" TID_b2c3d4e5f6a7b8c9 "-" "-" "-"
```

## Actual Behavior

The Lambda fails with:
```
failed to put object: failed to process: failed to decode: failed to decode text: invalid character 'x' in string escape code
```
Logs are not delivered to the destination.

## Root Cause Analysis

Three issues combine to produce this error:

### 1. `aws/v1` ELB preset regex does not match ALB log filenames

**`override/presets/aws/v1.yaml`**
```yaml
source: '\d{12}_elasticloadbalancing_[a-z\d-]+_[a-zA-Z0-9-]+_\d{8}T\d{4}Z_...'
#                                               ^^^^^^^^^^^^^^
#                                               Missing dot — fails to match "app.<alb-name>.<hexid>"
```

ALB log filenames embed the load balancer ARN as `app.<alb-name>.<hexid>` (with dots), but the character class `[a-zA-Z0-9-]` excludes dots. The preset **silently skips all ALB log files** and never sets `content-type: application/x-aws-elasticloadbalancing` or `content-encoding: gzip`.

### 2. Fallback decoding selects `TextDecoder` without a gzip wrapper

**`pkg/handler/forwarder/s3http/client.go` line 176**
```go
dec, err := decoders.Get(aws.ToString(params.ContentEncoding), aws.ToString(params.ContentType), params.Body)
```

With no matching preset, the file falls through to `infer/v1` or S3 object metadata, which may resolve to `content-type: text/plain` with no `content-encoding`. This causes `decoders.Get` to select `TextDecoder` with no gzip wrapper — meaning raw, undecompressed gzip bytes are fed directly to the decoder.

### 3. `TextDecoder` uses `strconv.Quote` which produces invalid JSON escapes

**`pkg/handler/forwarder/s3http/internal/decoders/text.go` lines 28–31**
```go
var buf bytes.Buffer
buf.WriteString(`{"text":` + strconv.Quote(s) + `}`)
if err := json.Unmarshal(buf.Bytes(), v); err != nil {
    return fmt.Errorf("failed to decode text: %w", err)
}
```

- `strconv.Quote()` uses **Go escape syntax** — for non-UTF-8 bytes (e.g., gzip magic bytes `0x1f 0x8b`) it produces `\x1f\x8b...`
- `json.Unmarshal()` follows **JSON spec (RFC 8259)** — `\x` is not valid JSON (only `\uXXXX` is)
- Result: `invalid character 'x' in string escape code` ❌

### Why it is silent for `s3://` destinations

**`pkg/lambda/forwarder/lambda.go` line 127**
```go
if strings.HasPrefix(cfg.DestinationURI, "https") {
    s3Client, err = s3http.New(...)  // decode pipeline only wired in for https
}
```

When `DESTINATION_URI` is `s3://`, the native AWS S3 client is used and `CopyObject` copies raw bytes directly — the decode pipeline (`decoders.Get`, `batch.Run`, `TextDecoder`) is **never invoked**, so the broken regex in `aws/v1.yaml` has no visible effect.

ddai1:
## Summary

Fixes two issues that together cause AWS ALB access log files to fail
processing when `DESTINATION_URI` is an HTTPS endpoint.

Closes #459

### 1. `pkg/handler/forwarder/override/presets/aws/v1.yaml`

Fix the `elasticLoadBalancing` preset regex to allow dots in the load
balancer name segment. ALB log filenames embed the ARN path as
`app.<alb-name>.<hexid>` (dots), but the previous character class
`[a-zA-Z0-9-]+` excluded dots, so the preset silently never matched
any ALB log file.

### 2. `pkg/handler/forwarder/s3http/client.go`

In `CopyObject`, infer `content-encoding: gzip` for `.gz` source keys
when `ContentEncoding` is not already set after override rules have run.

This handles the case where a `CONTENT_TYPE_OVERRIDES` rule sets
`content-type` (e.g. `application/x-aws-elasticloadbalancing`) but not
`content-encoding`. Because `Sets.Apply` stops at the first matching
set, `infer/v1`'s gzip rule never gets a chance to run — leaving
`ContentEncoding` empty. Without a gzip wrapper, the decoder receives
raw compressed bytes and produces garbage output or errors.

The check uses `getInput.Key` (the source object key) since the
destination key may differ and may not carry the `.gz` suffix.

## Testing

- Added `TestCopyObjectGzipInference` to `client_test.go`: creates a
  gzip-compressed ALB log body, mocks S3 `GetObject` returning it with
  `ContentType` set but `ContentEncoding: nil`, calls `CopyObject`, and
  asserts the HTTP destination receives decoded NDJSON (not raw binary).
- All existing tests pass: `TestClientPut`, `TestDecoders`.

### DIFF
--- a/pkg/handler/forwarder/override/presets/aws/v1.yaml
+++ b/pkg/handler/forwarder/override/presets/aws/v1.yaml
@@ -54,7 +54,7 @@
 
 - id: elasticLoadBalancing
   match:
-    source: '\d{12}_elasticloadbalancing_[a-z\d-]+_[a-zA-Z0-9-]+_\d{8}T\d{4}Z_[0-9.]+_[a-zA-Z0-9]+\.log\.gz$'
+    source: '\d{12}_elasticloadbalancing_[a-z\d-]+_[a-zA-Z0-9.-]+_\d{8}T\d{4}Z_[0-9.]+_[a-zA-Z0-9]+\.log\.gz$'
   override:
     content-type: 'application/x-aws-elasticloadbalancing'
     content-encoding: 'gzip'
--- a/pkg/handler/forwarder/s3http/client.go
+++ b/pkg/handler/forwarder/s3http/client.go
@@ -150,6 +150,14 @@ func (c *Client) CopyObject(ctx context.Context, params *s3.CopyObjectInput, opt
 
 	putInput := toPutInput(params, seekableBody, getResp.ContentType, getResp.ContentEncoding)
 
+	// Infer content-encoding for gzip files when not already set by override rules.
+	// This handles the case where a custom override sets content-type but not content-encoding,
+	// causing Sets.Apply to stop before infer/v1's gzip rule can run.
+	// We check the source key (getInput.Key) since the destination key may differ or lack the .gz suffix.
+	if aws.ToString(putInput.ContentEncoding) == "" && strings.HasSuffix(aws.ToString(getInput.Key), ".gz") {
+		putInput.ContentEncoding = aws.String("gzip")
+	}
+
 	putResp, err := c.PutObject(ctx, putInput, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to put object: %w", err)
--- a/pkg/handler/forwarder/s3http/client_test.go
+++ b/pkg/handler/forwarder/s3http/client_test.go
@@ -2,8 +2,10 @@ package s3http_test
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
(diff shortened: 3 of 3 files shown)


obs-gh-justindaines (approved):
Looks good. Thank you for contributing. I was able to test using filedrop and https.

PR #462, opened 2026-05-29
feat(metricsconfigurator): process FilterUri in the Lambda

obs-gh-virjramakrishnan:
## Summary

- Collapse the metricstream template's two filter paths (direct `AWS::CloudWatch::MetricStream` + `AWS::Include` macro vs. Lambda + GraphQL datasource) into a single Lambda-managed path. The MetricsConfigurator now downloads the `FilterUri` YAML from S3 itself and calls `PutMetricStream` directly.
- Drop the `MetricStream` resource and the `DeployLambda`/`NotDeployLambda` conditions; the Lambda is always deployed (FilterUri defaults to a non-empty observeinc-hosted file). `GQLTokenSecret` and the secrets-manager policy are gated on a new `HasDatasourceID` condition.
- Add `s3:GetObject` and `cloudwatch:DeleteMetricStream` to the Lambda role; pass `FilterUri` through the custom resource so file changes trigger an update; grant `lambda:*` on the CFN execution role in the metricstream integration test.
- Expose a new `MetricStreamNames` stack output containing the metric stream name prefix (`<StackName-or-NameOverride>-metric-stream-`) and a `aws cloudwatch list-metric-streams` snippet, since the streams are no longer visible as a CloudFormation resource in the stack's Resources tab.
- Accept "stream everything" filter files (e.g. `s3://observeinc/cloudwatchmetrics/filters/full.yaml`, which is `ExcludeFilters: []`). The original parser rejected this; now it parses, and the handler calls `PutMetricStream` once with no filters — CloudWatch's native way to express "all metrics in account."

This is the first half of splitting #458. PR 2 will follow with the SAM → plain CloudFormation conversion (`AWS::Lambda::Function` + S3 code parameters + the `embed-lambda-defaults.py` build pipeline).

## Customer-visible impact

- `FilterUri` parameter and the YAML schema (top-level `IncludeFilters` / `ExcludeFilters` arrays of `{Namespace, MetricNames}`) are unchanged — existing filter files (including `full.yaml`) remain compatible.
- For customers on the FilterUri-only path (no `DatasourceID`), the stack now also provisions the `MetricsConfigurator` Lambda, IAM role, log group, and a custom resource that previously only existed on the datasource path.
- The metric stream is no longer visible as a CloudFormation resource in the stack's Resources tab — it is created by the Lambda. The new `MetricStreamNames` output gives customers the name prefix needed to find streams in the CloudWatch console or via the CLI.
- Upgrade gap: on stack update, the CloudFormation-managed `MetricStream` is deleted and the Lambda re-creates it via `PutMetricStream`, with a brief window where metrics are not streamed.
- The metric stream is no longer drift-detected by CloudFormation (it is owned by a custom resource).

## Known issue → fast-follow PR

**Stream cleanup on stack delete is currently a no-op.** Pre-PR-1, the FilterUri path's metric stream was a CFN-managed resource and CFN deleted it cleanly. With this PR the stream is created by the Lambda, so on stack delete the stream is **orphaned** in CloudWatch (its Firehose target is destroyed by CFN at the same time, so the orphan delivers nothing — the operator must `aws cloudwatch delete-metric-stream` manually).

We deliberately do not auto-delete by name prefix: there's no way to prove a `<NameOverride>-metric-stream-N` stream was created by *this* stack vs. a sibling stack reusing the same `NameOverride`, a customer's hand-rolled stream, or a previous deploy in the same account+region. `DeleteMetricStream` also silently succeeds on nonexistent names, which made the previous prefix-loop run until Lambda timeout.

The follow-up PR will tag streams on `PutMetricStream` with the stack id and use `resourcegroupstaggingapi:GetResources` to find exactly this stack's streams on Delete. The IAM grant for `cloudwatch:DeleteMetricStream` is left in place here so the fast-follow is purely additive.

For the long-standing datasource path, this is no change: that path has had the orphan-on-delete bug since it was first added.

## Test plan

- [x] `make go-test` passes.
- [x] `make go-lint` passes (0 issues).
- [x] `make sam-validate` passes for all apps including metricstream.
- [x] Manual CRUD walk-through against blunderdome / us-west-2 with the composite stack and FilterUri = `recommended.yaml`:
  - **Create** — stack reaches `CREATE_COMPLETE`; `MetricStreamNames` output reads `<StackName>-MetricStream-metric-stream-`; the named stream is `running` with the expected `ExcludeFilters` populated; Lambda log shows the FilterUri path and `Status: SUCCESS`.
  - **Update** (after pushing the `full.yaml` regression fix) — switch FilterUri to `full.yaml`; Lambda log shows `using FilterUri path` with the new URI; the stream's `LastUpdateDate` advances and both filter lists are now `null` (= "stream everything"); stack reaches `UPDATE_COMPLETE`.
  - **Delete** — stack reaches `DELETE_COMPLETE`; with the no-op `handleDelete`, the metric stream is left in CloudWatch as expected (the operator manually deleted it after the test).
- [ ] CI: integration test `metricstream` (validates the FilterUri-only path end-to-end with the new Lambda).
- [ ] CI: integration test `stack` (composite — covers the MetricStream substack via the parent template).


apps/metricstream/template.yaml
```diff
@@ -99,12 +99,10 @@ Conditions:
   UseStackName: !Equals
     - !Ref NameOverride
     - ''
-  DeployLambda: !Not
```
obs-gh-virjramakrishnan, line 102: Since the lambda is unconditionally deployed, no need for this variable any more.

apps/metricstream/template.yaml
```diff
@@ -99,12 +99,10 @@ Conditions:
...
+  HasDatasourceID: !Not
```
obs-gh-virjramakrishnan, line 102: now, the relevant question to ask is whether the user included a DatasourceID. In that case, we should read from the datastream to get the configuration, as opposed to reading from the s3 file.

apps/metricstream/template.yaml
```diff
@@ -230,33 +228,8 @@ Resources:
                   - firehose:PutRecord
                   - firehose:PutRecordBatch
                 Resource: !GetAtt 'DeliveryStream.Arn'
-  MetricStream:
```
obs-gh-virjramakrishnan, line 233: We no longer want to conditionally deploy the metric stream as part of the stack: the lambda will create it. Of course, this means that the metric stream won't show up as part of the stack, but it should still be accessible in the AWS console if the user wants to look. To this end, I've added an output variable that should point to the metric stream(s) created, so users can see the streams they've created.

apps/metricstream/template.yaml
```diff
@@ -310,32 +294,35 @@ Resources:
...
+            - ''
   GQLTokenSecret:
     Type: AWS::SecretsManager::Secret
-    Condition: DeployLambda
+    Condition: HasDatasourceID
     Properties:
       Description: GQL Token Secret
       Name: !Sub "observe-gql-token-${AWS::StackName}"
       SecretString: !Ref GQLToken
   StackCreationUpdateCustomResource:
```
obs-gh-virjramakrishnan, line 319: This resource's purpose is to trigger the lambda if something changes in the stack. If the customer changes the FilterURI, it will trigger the stack to update (we need it to, because we need the lambda to run again).

obs-gh-virjramakrishnan (commented):
This change should just switch over s3 file processing to the lambda. This should be mostly invisible to the customer, the fields they enter are the same, only the inner process changes.

pkg/handler/metricsconfigurator/filteruri.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	path := strings.TrimPrefix(uri, "s3://")
+	parts := strings.SplitN(path, "/", 2)
+	if len(parts) < 2 || parts[1] == "" {
+		return "", "", fmt.Errorf("invalid S3 URI: must contain bucket and key")
+	}
+	return parts[0], parts[1], nil
+}
+
+func downloadFilterYAML(ctx context.Context, cfg aws.Config, filterURI string) ([]byte, error) {
```
obs-gh-mattcotter: This function looks a lot like `downloadPollerConfig`. Can we share code between the two?

pkg/handler/metricsconfigurator/filteruri.go
```diff
@@ -0,0 +1,102 @@
...
+	ExcludeFilters []types.MetricStreamFilter
+}
+
+func parseFilterYAML(data []byte) (*ParsedFilters, error) {
+	var raw FilterYAML
+	if err := yaml.Unmarshal(data, &raw); err != nil {
+		return nil, fmt.Errorf("failed to parse filter YAML: %w", err)
+	}
+
+	// Distinguish "key present but empty list" (e.g. `ExcludeFilters: []` in
```
obs-gh-mattcotter: Not sure if this works, but can we use the type `*[]FilterEntry` in `FilterYAML` to handle this, then check for null?

pkg/handler/metricsconfigurator/handler.go
```diff
@@ -85,64 +87,75 @@ func New(cfg *Config, logger logr.Logger) (Handler, error) {
...
 	}
-	logger.V(4).Info("parsed response, metric filters", "filters", MetricsFilters)
+	logger.V(4).Info("parsed response, metric filters", "filters", metricsFilters)
 
-	// Create a cloudwatch client
 	cwClient := cloudwatch.NewFromConfig(cfg)
 
-	// AWS limits the number of metrics in a metric stream to 1000
-	// https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricStreamFilter.html
-	// Make multiple metric streams to account for this
```
obs-gh-mattcotter, line 143: This comment got lost

pkg/handler/metricsconfigurator/filteruri_test.go
```diff
@@ -0,0 +1,235 @@
...
+			wantIncludes: 2,
+			wantExcludes: 0,
+		},
+		{
+			name: "include filters without metric names",
+			yaml: `IncludeFilters:
+  - Namespace: AWS/EC2
+  - Namespace: AWS/EBS
+`,
+			wantIncludes: 2,
```
obs-gh-mattcotter, line 76: I think we're missing a test where both include and exclude are specified. From the handler, it looks like this case isn't supported, but I don't think it will error.

pkg/handler/metricsconfigurator/filteruri.go
obs-gh-virjramakrishnan: great point! I factored out the code so that there's one common function that does s -> byte[] and then the poller/metricStream implementations take it from there

pkg/handler/metricsconfigurator/filteruri.go
obs-gh-virjramakrishnan: yes, in fact we can unmarshal directly to a `[]FilterEntry`, no need for the pointer. If it is provided and empty, it will show up as `[]` and if it is not provided, it will show up as `null`.

pkg/handler/metricsconfigurator/filteruri_test.go
obs-gh-virjramakrishnan: I think AWS disallows this [explicitly](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricStream.html#API_PutMetricStream_RequestParameters). I could run a test to check that it errors though! I've added a line in the readme doc for metricstream to make this limitation clear.

pkg/handler/metricsconfigurator/filteruri_test.go
```diff
@@ -0,0 +1,181 @@
...
+			wantExcludes: 0,
+		},
+		{
+			name:         "yaml with neither field",
+			yaml:         `SomethingElse: value`,
+			wantIncludes: 0,
+			wantExcludes: 0,
+		},
+		{
+			name: "both IncludeFilters and ExcludeFilters non-empty rejected",
```
obs-gh-mattcotter, line 97: Can you add a test for:

```yaml
IncludeFilters: []
ExcludeFilters: []
```
? I think this will not error as written. Up to you whether or not we should error on that, but we should test it and leave a comment either way.

obs-gh-mattcotter (approved):
One more test / comment request, otherwise lgtm!

pkg/handler/metricsconfigurator/filteruri_test.go
obs-gh-virjramakrishnan: thanks for this case, I didn't consider that! I will just check what AWS would do (I think it would error) and conform to that.

obs-gh-virjramakrishnan: actually, turns out it does not error. AWS treats `[]` as nothing. Added the test!

obs-gh-mattcotter (approved):

PR #463, opened 2026-05-30
feat(metricsconfigurator): tag-based metric stream cleanup on stack delete

obs-gh-virjramakrishnan:
## Summary

Fast-follow to #462. PR 1 left \`handleDelete\` as a no-op; this PR makes stack delete actually clean up the metric streams it created, with ownership semantics that don't risk deleting other stacks' or customer-owned streams.

- On every \`PutMetricStream\` (both \`invokeFilterUriPath\` and \`invokeDatasourcePath\`), pass \`Tags=[{aws-sam-apps:stack-id: req.StackId}]\`. \`req.StackId\` is already in the custom-resource payload — no extra state to persist.
- \`handleDelete\` now pages through \`cloudwatch:ListMetricStreams\` for the account, calls \`ListTagsForResource\` on each, and \`DeleteMetricStream\` only on streams whose \`aws-sam-apps:stack-id\` tag matches the current stack's id.
- IAM additions on the Lambda role: \`cloudwatch:TagResource\` (required when \`PutMetricStream\` is called with \`Tags\`), \`cloudwatch:ListTagsForResource\`, and \`cloudwatch:ListMetricStreams\` (account-wide; no per-resource ARN).
- Lambda \`Timeout\` bumped from SAM's 3-second default to 60s — \`ListMetricStreams\` + N×\`ListTagsForResource\` is comfortably slower than that with the old 3s ceiling.

## Why this exists

Discovered while doing the CRUD walk-through on #462: the original cleanup loop walked stream names by index calling \`DeleteMetricStream\`, but \`DeleteMetricStream\` silently succeeds on nonexistent names, so the loop ran until Lambda timeout. Worse, a prefix-match by \`MetricStreamName\` would happily delete streams owned by *another* stack reusing the same \`NameOverride\`, customer-managed streams that happen to match the prefix, or streams from a previous deploy in the same account+region.

We deliberately scoped the fix to a separate PR so #462 could land first with the cleanup deferred (no-op + a documented \"orphan on delete\" caveat).

## Side benefit

The pre-existing **datasource** path has had the orphan-on-delete bug since it was first added — its original \`Delete\` handler just returned SUCCESS without calling \`DeleteMetricStream\`. Tagging on \`PutMetricStream\` + tag-filter cleanup covers both paths uniformly, so this also retroactively fixes that.

## Customer-visible impact

- On stack delete, the metric stream(s) the Lambda created are removed automatically. No more orphans (no manual \`aws cloudwatch delete-metric-stream\` cleanup, no surprise CloudWatch metric-stream charges after delete).
- New IAM grants on the Lambda role's policy: \`cloudwatch:TagResource\`, \`cloudwatch:ListTagsForResource\`, \`cloudwatch:ListMetricStreams\`.
- Streams created by previous versions of this stack (before this PR landed) **will not** be cleaned up on delete — they don't have the tag, and we will not touch streams whose ownership we can't prove. Operators upgrading from #462 should manually clean up any leftover streams from prior deploys.

## Test plan

- [x] \`make go-test\` passes.
- [x] \`make go-lint\` passes (0 issues).
- [x] \`make sam-validate\` passes.
- [ ] Manual CRUD walk-through against blunderdome / us-west-2 (planned, similar to #462).
- [ ] CI: integration test \`metricstream\`.
- [ ] CI: integration test \`stack\`.

## Stacking

This PR targets \`feat/metricsconfigurator-filteruri\` (#462). Once #462 merges to main, GitHub will retarget this PR to main.

### DIFF
--- a/apps/metricstream/template.yaml
+++ b/apps/metricstream/template.yaml
@@ -247,8 +247,14 @@ Resources:
                 Action:
                   - cloudwatch:PutMetricStream
                   - cloudwatch:DeleteMetricStream
+                  - cloudwatch:TagResource
+                  - cloudwatch:ListTagsForResource
                 Resource: !Sub |-
                   arn:aws:cloudwatch:${AWS::Region}:${AWS::AccountId}:*
+              - Effect: Allow
+                Action:
+                  - cloudwatch:ListMetricStreams
+                Resource: '*'
               - Effect: Allow
                 Action:
                   - iam:PassRole
@@ -287,6 +293,7 @@ Resources:
       CodeUri: ../../bin/linux_arm64
       Handler: bootstrap
       Runtime: provided.al2023
+      Timeout: 60
       Architectures:
         - arm64
       LoggingConfig:
--- a/pkg/handler/metricsconfigurator/handler.go
+++ b/pkg/handler/metricsconfigurator/handler.go
@@ -32,6 +32,15 @@ func stackIdTags(stackId string) []types.Tag {
 	}
 }
 
+// cwAPI is the subset of cloudwatch.Client this handler uses. Defined as
+// an interface so tests can swap in a fake without touching real AWS.
+type cwAPI interface {
+	PutMetricStream(ctx context.Context, input *cloudwatch.PutMetricStreamInput, optFns ...func(*cloudwatch.Options)) (*cloudwatch.PutMetricStreamOutput, error)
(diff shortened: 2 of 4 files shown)


obs-gh-justindaines (approved):
Looks good, and I see https://github.com/observeinc/aws-sam-apps/pull/462 was merged.

PR #464, opened 2026-06-02
feat: convert logwriter/metricstream/externalrole to plain CloudFormation

obs-gh-virjramakrishnan:
## Summary

Second half of splitting #458. PR #462 landed the FilterUri-via-Lambda change; this PR is the SAM → plain CloudFormation conversion that needed to follow because **CloudFormation StackSets do not support the SAM transform**.

- Switch \`logwriter\` (subscriber), \`metricstream\` (metricsconfigurator), and \`externalrole\` (pollerconfigurator) templates from \`AWS::Serverless::Function\` to \`AWS::Lambda::Function\`. Lambda code is referenced via \`Code: { S3Bucket, S3Key }\` parameters (\`LambdaS3BucketPrefix\` / \`LambdaS3Key\`) instead of inline \`CodeUri\`.
- Build pipeline: \`subscriber\`, \`metricsconfigurator\`, and \`pollerconfigurator\` get zipped into Lambda code archives (\`.aws-sam/build/lambda-zips/\`) and uploaded to S3 during \`sam-package\`. \`scripts/embed-lambda-defaults.py\` post-processes the packaged template and writes \`LambdaS3BucketPrefix\` / \`LambdaS3Key\` defaults so the published artifact is self-contained.
- Apps split into \`SAM_APPS\` / \`CFN_APPS\` / \`STACKSET_APPS\` in \`variables.mk\`. SAM apps validated with \`sam validate\`; CFN apps validated with \`yamllint\` + \`cfn-lint\`. \`PACKAGEABLE_APPS = SAM_APPS + CFN_APPS\`.
- The \`stack\` composite template plumbs \`LambdaS3BucketPrefix\` / \`LambdaS3Key\` parameters through to its nested CFN substacks.
- CI: \`tests.yaml\` installs \`cfn-lint\` + \`yamllint\` (not pre-installed on GH runners).
- Two CFN apps suppress the \`W1031\` cfn-lint check on the \`!Sub '\${LambdaS3BucketPrefix}-\${AWS::Region}'\` bucket reference — \`LambdaS3BucketPrefix\` defaults to empty for cfn-lint's purposes; \`embed-lambda-defaults.py\` writes the real value at package time. Runtime templates always have a real prefix.
- \`metricstream\` template loses a now-dead \`DeployLambda\` condition — PR #462 already collapsed the dual filter paths, so the Lambda is unconditionally deployed.

## Why this is a separate PR from #462

The two changes touched the same files but had distinct rationales: #462 was a behavior change (FilterUri now goes through the Lambda instead of \`AWS::Include\`), this is structural (transform method). Splitting kept #462 reviewable and let #463 (the tag-based cleanup fast-follow) stack on a stable base.

## Customer-visible impact (upgrade path)

- New required-at-package-time parameters: \`LambdaS3BucketPrefix\`, \`LambdaS3Key\`. Customers using the published Observe template never set these manually — they're written as defaults during \`sam-package\` by the embed script. Customers running \`sam deploy\` themselves need to supply them.
- Resource-type swap: \`AWS::Serverless::Function\` → \`AWS::Lambda::Function\`. CloudFormation treats this as a logical-id-stable in-place update for the function itself, but **event source mappings hanging off the function get new logical IDs and may be replaced** during the upgrade window. See the "Upgrading from SAM-based templates" section of \`DEVELOPER.md\`.
- The \`metricstream\` substack's metric stream is created by the Lambda (already true after #462), so no additional gap from this PR specifically.

## Test plan

- [x] \`go-test\`, \`go vet\`, \`make sam-validate\` all pass locally on the rebased branch.
- [x] \`cfn-lint\` clean on \`logwriter\`, \`metricstream\`, \`externalrole\` templates.
- [ ] CI: integration tests \`logwriter\`, \`metricstream\`, \`externalrole\`, and \`stack\` (composite — exercises all three substacks).


apps/externalrole/template.yaml
```diff
@@ -79,8 +162,130 @@ Resources:
...
+            Action: 'sts:AssumeRole'
+      Policies:
+        - PolicyName: PollerConfiguratorPolicy
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - logs:CreateLogStream
+                  - logs:PutLogEvents
```
obs-gh-mattcotter, line 218: Does this need `s3:GetObject`?

docs/stack.md
```diff
@@ -6,65 +6,73 @@ The Observe AWS stack is designed to aggregate and forward a wide range of AWS r
 
 This stack centralizes the collection process by deploying necessary AWS services and linking them to ensure seamless data flow towards your Observe Filedrop.
 
-![Stack](images/stack.png)
```
obs-gh-mattcotter, line 9: Did you mean to remove this image?

scripts/embed-lambda-defaults.py
```diff
@@ -0,0 +1,79 @@
...
+                r"(?:    .*\n)*?"
+                r'    Default: )"[^"]*"',
+                re.MULTILINE,
+            )
+            replacement_dq = r'\g<1>"' + param_value + '"'
+            new_content = pattern_dq.sub(replacement_dq, content)
+
+        if new_content == content:
+            # Try unquoted empty default (Default: '')
+            # Some SAM-packaged templates may use Default: '' on the same line
```
obs-gh-mattcotter: Is this comment old? Should there be some error log here?

docs/stack.md
obs-gh-virjramakrishnan: No - sorry about this typo.

scripts/embed-lambda-defaults.py
obs-gh-virjramakrishnan: Yes, it is dead code. I think I incorrectly split up the commits of the original stackset PR. I'll square this away.

apps/externalrole/template.yaml
obs-gh-virjramakrishnan: Yes it does. Thank you for catching this! I think this got improperly split up when I split up the stacksets PR .

obs-gh-mattcotter (approved):

PR #465, opened 2026-06-09
feat: add StackSet wrappers for org-wide deployment

obs-gh-virjramakrishnan:
## Summary

Stacked on top of #464 (plain-CFN conversion). Adds CloudFormation StackSet wrappers so the logwriter / metricstream / externalrole apps can be deployed across an AWS Organization from a single template.

- **3 stackset wrappers** (`apps/<app>-stackset/template.yaml`) — each wraps the underlying CFN_APP via `AWS::CloudFormation::StackSet` with `SERVICE_MANAGED` permissions, auto-deployment, and parameterised `OperationPreferences` (defaults: 100% concurrent, 100% failure tolerance, parallel regions).
- **Build** — `sam-package` now embeds a region-specific `TemplateURL` default in each stackset wrapper so customers don't have to construct the underlying app's S3 URL. `embed-lambda-defaults.py` gains an "insert Default if missing" branch (TemplateURL has no source-side default). `sam-pull` now downloads stackset templates too.
- **Docs** — `docs/multi-account.md` covers prereqs, central bucket permissions, concurrency tuning, monitoring, partial-failure recovery, and cleanup. DEVELOPER.md gets a StackSet workflow section.

## Test plan

- [ ] CI: `cfn-lint` clean on all 3 stackset wrappers, `yamllint` clean
- [ ] `make sam-package-<app>` produces `<app>-stackset.yaml` with `TemplateURL` default pointing at the just-uploaded `<app>.yaml` in the same bucket+version (verified locally for externalrole)
- [ ] One end-to-end stackset deploy in a test org once #464 lands

## Stacking

Once #464 merges, GitHub will auto-rebase this PR onto main.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -35,17 +35,18 @@ and the most important variables, run `make help`, e.g:
 
 ```
 VARIABLES:
-  APPS          = config configsubscription firehose forwarder logwriter metricstream stack
+  APPS          = config configsubscription externalrole externalrole-stackset forwarder logwriter logwriter-stackset metricstream metricstream-stackset stack
   AWS_REGION    = us-west-2
-  GO_BINS       = forwarder subscriber metricsconfigurator
+  GO_BINS       = forwarder subscriber metricsconfigurator pollerconfigurator
   GO_BUILD_DIRS = bin/linux_arm64 .go/bin/linux_arm64 .go/cache .go/pkg
-  TF_TESTS      = config configsubscription firehose forwarder forwarder_s3 logwriter metricstream simple stack
+  TF_TESTS      = config configsubscription externalrole forwarder forwarder_s3 logwriter metricstream simple stack stack_including_metricspollerrole
   VERSION       = v1.19.2-4-gb1238b5-dirty
 
 TARGETS:
   clean                           removes built binaries and temporary files.
   go-build                        build Go binaries.
   go-clean                        clean Go temp files.
+  lambda-zips                     build Lambda ZIP archives for non-SAM-managed functions.
   ...
 ```
 
@@ -57,14 +58,26 @@ TARGETS:
 
 # package a single SAM app, will upload lambda function to S3
 → make sam-package-forwarder
+→ make sam-package-logwriter
 ```
 
 At this point, you will have a functional CloudFormation template under
-`.aws-sam/build/regions/${AWS_REGION}/forwarder.yaml`. You can deploy this by:
+`.aws-sam/build/regions/${AWS_REGION}/${APP}.yaml`. You can deploy this by:
 
 - uploading manually using Cloudformation console
 - deploying through `sam deploy`
 - installing via Terraform
+
+For apps with Lambda functions that SAM doesn't manage natively (logwriter,
+metricstream, externalrole, stack), the packaging step automatically builds
+Lambda ZIPs, uploads them to S3, and embeds the S3 references as parameter
+defaults in the packaged template. An S3 bucket is auto-created using the
+naming convention `aws-sam-apps-${ACCOUNT_ID}-${REGION}`. You can override
+this by setting `S3_BUCKET_PREFIX`:
+
+```
+S3_BUCKET_PREFIX=my-bucket- make sam-package-logwriter
+```
 
 ## Running tests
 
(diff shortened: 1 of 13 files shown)


apps/externalrole-stackset/template.yaml
```diff
@@ -0,0 +1,247 @@
...
+        it for metrics polling, with optional poller configuration.
+      TemplateURL: !Ref TemplateURL
+      PermissionModel: SERVICE_MANAGED
+      CallAs: !Ref CallAs
+      AutoDeployment:
+        Enabled: !Ref AutoDeployEnabled
+        RetainStacksOnAccountRemoval: !Ref RetainStacksOnAccountRemoval
+      Capabilities:
+        - CAPABILITY_IAM
+        - CAPABILITY_NAMED_IAM
```
obs-gh-mattcotter, line 205: Do these templates need `CAPABILITY_AUTO_EXPAND`? It's listed in the docs.

apps/logwriter-stackset/template.yaml
```diff
@@ -0,0 +1,303 @@
...
+    MaxValue: 64
+    Description: >-
+      Buffer incoming data to the specified size, in MiBs, before delivering
+      it to S3.
+  NumWorkers:
+    Type: String
+    Description: >-
+      Maximum number of concurrent workers when processing log groups.
+    Default: ''
+  MemorySize:
```
obs-gh-mattcotter, line 210: A lot of the params in the stackset templates have dropped the `AllowedValues` / `AllowedPattern` present in the underlying template. This isn't true for every param though (ex: `UpdateTimestamp` in the external role).

My suggestion is to keep them in both places, and add a test script to parse the yamls and validate that the params match. What do you think?

DEVELOPER.md
```diff
@@ -35,17 +35,18 @@ and the most important variables, run `make help`, e.g:
 
 ```
 VARIABLES:
-  APPS          = config configsubscription firehose forwarder logwriter metricstream stack
```
obs-gh-mattcotter, line 38: What was `firehose`? Something that was removed before your current change?

obs-gh-virjramakrishnan: This was a legacy lambda, long since gone. See here: https://github.com/observeinc/aws-sam-apps/pull/309/changes. This is cleanup my agent found - let me know if it's better placed in a separate PR!

apps/externalrole-stackset/template.yaml
obs-gh-virjramakrishnan: The docs are out of date: all apps that now use plain cfn no longer need this permission. It's mentioned in the docs because it's still necessary for the stack called `stack`, which is still SAM

apps/logwriter-stackset/template.yaml
obs-gh-virjramakrishnan: Very fair on the test script part! I will add that in.

DEVELOPER.md
obs-gh-mattcotter: Thanks for cleaning that up, no need to separate it out

scripts/check-stackset-param-parity.py
```diff
@@ -0,0 +1,158 @@
...
+    """Extract a single constraint's value from a parameter block.
+    Returns None if absent. Inline values are returned as a stripped
+    string. List values (e.g. AllowedValues across several lines) are
+    normalized to a sorted bracketed string for comparison."""
+    # Try multi-line list form first (must come before the inline check
+    # because both can match an empty inline value):
+    #     AllowedValues:
+    #       - 'foo'
+    #       - 'bar'
+    list_match = re.search(
```
obs-gh-mattcotter, line 89: I think that both this and `parse_param_blocks` would be a lot cleaner if they used a yaml library (ex: [PyYAML](https://pyyaml.org/wiki/PyYAMLDocumentation)).

scripts/check-stackset-param-parity.py
obs-gh-mattcotter, line 1: How much effort would it be to add this to CI? I imagine this runs very quickly, so I think it's worth doing to prevent us forgetting in the future.

obs-gh-mattcotter (approved):
I left a few comments on the test script, but the rest look great, thank you! Feel free to either merge after addressing the testing comments or address them in a follow up.

scripts/check-stackset-param-parity.py
obs-gh-virjramakrishnan: ah, the issue here is that the SAM app formatting messes with PyYAML. Tell you what, I'll see if I can get this to work with some customization of PyYAML as opposed to a fully custom solution

obs-gh-mattcotter: Oh I see; if the files aren't valid yaml don't worry about it.

scripts/check-stackset-param-parity.py
obs-gh-virjramakrishnan: This is already sneakily in CI: its a prereq of `sam validate`, so it will cause that CI test to fail if it fails.

PR #469, opened 2026-06-18
feat: allow pollerconfigurator lambda to create a poller with resource tagging enabled

obs-gh-mattcotter:


obs-gh-virjramakrishnan (approved):
LGTM!

